PR #24, opened 2014-03-24
revert recent bytes/str commit: fails for binary data

geggo:
Dear Hernan,

with the recent commit f7f06d5 I experienced problems with reading binary data. In particular, '.decode('ascii')' in from_string_buffer_raw() fails for binary data containing characters > 127. I think the result from a read command needs to be a bytes, not a string object, for not loosing any data. I propose to simply revert the commit.

This problem was masked by catching and ignoring all exceptions in read_raw()

Gregor


arturaugusto:
Still having this issue when I read a instrument that returns the '@' character. 
I'm using python 2.7.

PR #56, opened 2014-07-10
Added type checking to parse_resource_extended

Joshuacru:
Added type checking to parse_resource_extended to prevent segmentation fault caused when resource_name is an integer.


hgrecco:
I think we do not need this

PR #114, opened 2015-01-21
Fixed access to the interface_type property

mordoror:
When trying to access the 'interface_type' property of my opened resource, I got the following error:
```  File "C:\Python27\lib\site-packages\pyvisa\resources\resource.py", line 145, in interface_type
    self.resource_name).interface_type
AttributeError: 'tuple' object has no attribute 'interface_type' ```

Indexing the first element of that tuple fixed the problem.


hgrecco:
Thanks!

PR #139, opened 2015-04-23
Improve handlers

lupien:
This is my first pull request on GitHub so pardon me if it is not done properly.

These are a list of commits to fix/improve the usage of events/handlers


hgrecco:
Great PR. Merged! 

One more thing, how about simplifying the uninstall? Is it possible?

PR #143, opened 2015-04-28
Compat struct

daritter:
create compact/struct.py to replace unicode literals with str() on python<2.7.8 and modify util.py to use this


hgrecco:
Thanks!

PR #156, opened 2015-05-13
Moved control_ren to a mixin and added it to USB, GPIB and TCPIP

lupien:
control_ren can be used by more than gpib (can change local/remote state of instrument on usb and tcpip). I created a mixin and used it where I know it will work.
Another option would be to just put control_ren in MessageBasedResource.


hgrecco:
Merged Thanks!

PR #189, opened 2015-10-07
dedup and change instrument classes

JMoVS:
it previously stated RegisterBasedResource twice, which didn't make sense and later referenced MessageBasedResource which wasn't mentioned before.


JMoVS:
I have no idea why a simple word change would fail compilation.
Is my assumption correct that the documentation had errors at this place?

MatthieuDartiailh:
The error is unrelated to your change. It seems that coverage (used to run the testsuite) does not support Python 3.2, hence the failure. @hgrecco should python 3.2 support be dropped ? (issue with u'')

hgrecco:
@MatthieuDartiailh It is actually coverage which dropped Python 3.2. I will just force coverage 3.7 until w decide to do it.

PR #191, opened 2015-10-09
delayed forced read

PhaseShifter:
Just wait read_delay amount of time, then fetch all the data

This seems to work for me, but I have no idea of how pythonic this is. Git and github are also quite new for me, so I'm open for constructive criticism ;-)

### DIFF
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -88,7 +88,6 @@ class MessageBasedResource(Resource):
 
     chunk_size = 20 * 1024
     query_delay = 0.0
-    read_delay = 0.1
 
     _values_format = None
 
@@ -300,7 +299,7 @@ class MessageBasedResource(Resource):
         ret = bytes()
 
         if self._read_termination is None:
-            time.sleep(self.read_delay)
+            time.sleep(self.query_delay)
             chunk, status = self.visalib.read(self.session, self.bytes_in_buffer)
             ret += chunk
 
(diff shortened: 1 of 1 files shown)


hgrecco:
Thanks for the PR. I

Two comments about the proposed implementation. `read_delay` as an instance attribute is not a good idea. There is no (common) scenario in which you want a `read_delay`. What you want is a delay in  query, between the read and write operation. And that option already exists as `query_delay` (which can be given as a parameter to query defaulting to an instance attribute).

I found interesting the other (implicit) proposal of your code. You propose that `read_termination = None` is equated to read everything from the buffer. I think is an interesting idea and should not break any existing code. But we need to careful considere this.

PhaseShifter:
Thanks for the feedback! I agree `query_delay` is the attribute to use instead of `read_delay`, I tried to assign a value to that variable in the first place, in combination with the `read_termination = None` condition, and expected it to work as it does in this branch.

About my proposal of reading everything in the buffer when `read_termination` equates to None.  
I wonder what other conditions might apply when that is the case. Maybe the termination would be a fixed number of bytes in the buffer? These conditions could be combined (return the acquired data when X bytes have been received, or when X time has passed). That would speed up the program when you know how much data to expect. What other conditions could apply?

hgrecco:
A fixed number of bytes is not the common case for message based communication. Most of the times messages are flexible and there is a way to determine when the message is over. I do agree that the current behavior is not very convenient for fixed size messages. 

I see several roadblocks that we need to overcome:
1. The value `None` means "remove termination" when used for `read_termination` but "use the default termination" when used for the termination argument of read.
2. There is a `size` argument, but it only refers to the chunk size not the message size.

MatthieuDartiailh:
Closed by the introduction of read_bytes in #301

Issue #228, opened 2016-07-12
VISA Shell fails on list command with resources available


Roanmh:
Upon executing the list command, an uncaught exception is thrown, crashing the shell. See below for a list of resources gained through programmatic use of PyVISA. The Error and stack trace is shown below. The environment is Windows NT4.0 (outdated yes, but I am submitting in-case bug also effects other systems) running Python 2.6 and PyVISA 1.8. I can also provide more information upon request.

**Custom program showing current available resources:**
![image](https://cloud.githubusercontent.com/assets/10479270/16784036/e1054158-4854-11e6-9089-6be090b7518d.png)

---

**Failing run of VISA Shell (also shows a degree of pwd independence):**
![image](https://cloud.githubusercontent.com/assets/10479270/16784081/21dace78-4855-11e6-8dab-d7da6b258f0d.png)

---

**Additionally, use of the open command with an explicit resource name gives the message of the same error (without a fatal failure, however):**
![image](https://cloud.githubusercontent.com/assets/10479270/16784153/6c443a3a-4855-11e6-9d51-4408f18dc13c.png)

---

**Note:** If anyone has insight as to this issue, I would be very interested to hear it as I may also put some work into this bug, myself.

Roanmh:
After some research I found that this a compatibility issue with Python 2.6. Python 2.6 requires that format strings have indexes in variable placeholders. e.g. `{1}` works, but `{}` will not in python 2.6. Since PyVISA is listed as being compatible with python 2.6, it seems like this is a bug within pyVISA area of support.

@hgrecco Are you interested in fixing this or reviewing a pull request for this issue if I am able to?

hgrecco:
Fixed now. Thanks!

PR #233, opened 2016-08-05
Added explicit numbering to format strings for py2.6 compatibility.

Roanmh:
In response to issue #228, this pull request addresses the need for explicit naming/numbering of format tags in the format strings for 2.6 compatibility.

--

Additionally, help with testing would be much appreciated as I do not have frequent access to a system for such testing or access to environments with later versions and visa libraries. With guidance (on setting up an environment to use the module), I may be able to test more thoroughly as well. Any tips or corroborative testing are much appreciated.


hgrecco:
Merged. Thanks!

PR #238, opened 2016-10-11
fix(highlevel): added resource_name to parse_resource_extended

famish99:
Using ResourceManager.resource_info returned inconsistent results
between the NI-VISA backend vs other backends which didn't extend
visalib.parse_resource_extended (e.g. PyVISA-sim) and code relying
on the "normalization" behavior of resource_info.resource_name
would not work.


MatthieuDartiailh:
This looks good to me. bors r+

PR #249, opened 2016-12-09
Enable Exceptions to be pickled/unpickled

bernizt:
Fixes a bug when trying to pickle/unpickle pyvisa exceptions. For example, previously it was not possible to send an exception instance from one process to another through a pipe.

### DIFF
--- a/pyvisa/errors.py
+++ b/pyvisa/errors.py
@@ -368,6 +368,9 @@ class VisaIOError(Error):
         self.error_code = error_code
         self.abbreviation = abbreviation
         self.description = description
+        
+    def __reduce__(self):
+        return (VisaIOError, (self.error_code,))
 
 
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Could you please add a test ? Once a test has been added I will try to merge quickly.

bernizt:
Of course. I hope these new tests are enough.

MatthieuDartiailh:
LGTM. At one point Github displayed an added file but it is gone now. @hgrecco @thliebig any comments ?
Otherwise I will merge in about a week. Thanks @bernizt

bernizt:
I uploaded the file in a wrong directory so I had to amend it afterwards, so that is probably why an added file was shown for some minutes.

MatthieuDartiailh:
Thanks @bernizt

PR #255, opened 2017-01-24
Turned Resource into a context manager

madphysicist:
This makes it possible to do `with rm.open_resource('...') as res: ...`.

### DIFF
--- a/pyvisa/resources/resource.py
+++ b/pyvisa/resources/resource.py
@@ -113,6 +113,12 @@ class Resource(object):
     def __repr__(self):
         return "<%r(%r)>" % (self.__class__.__name__, self.resource_name)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
     @property
     def last_status(self):
         """Last status code for this session.


MatthieuDartiailh:
@hgrecco, @thliebig any opinion on this ? It seems quite straightforward but I am not sure we want to see the ResourceManager grow new methods.
@madphysicist what is your rationale behind this change ?

madphysicist:
I have come to expect the convenience of context management from any Python object that does error-prone I/O. It is much easier to deal with a `with` block than a `try-except-finally`. I think that this is true for the average Python user in general. I pretty much extend or wrap all my resources to enable this functionality at the moment, so I figured that adding context management at the lowest possible level would make people's lives a bit easier.

hgrecco:
I use contextmanagers as an alternative to initialize methods in lantz and works quite well. But I think it only makes sense for short code.

MatthieuDartiailh:
If the ResourceManager was able to cleanly close any resources left open, it may be a worthy addition as all the code could below the `with ResourceManager() as rm:` no matter the number of instruments used. I really think we should add such capabilities to the ResourceManager and a nice error in resources trying to used a closed resource manager.

madphysicist:
@MatthieuDartiailh That might make more sense.

MatthieuDartiailh:
Actually both can cohabitate. But I had misread your patch and though it applied to the ResourceManager already, hence my previous comments.
So this is good to go for me. I will wait a bit to see if anybody else comment and otherwise I will merge. Do not hesitate to ping me if I forget.

MatthieuDartiailh:
By the way @madphysicist would you be interested in making the ResourceManager usable as a context too (which would require adding a tracking of the opened resources) ?

madphysicist:
@MatthieuDartiailh. Sure. Would you like a separate PR for that?

MatthieuDartiailh:
Yes a separate PR would be better since we need to add the tracking of created resources to the manager. I mentioned a WeakSet before but I actually think that a WeakValueDict whose key would be the normalized visa resource string may prove more appropriate since it will allow to query the opened resources (which is something that popped up recently in an issue).

madphysicist:
`WeakValueDict` sounds like a reasonable choice. I will start something tonight. Should have a PR ready in a day or two.

MatthieuDartiailh:
Sounds great

MatthieuDartiailh:
@madphysicist any progress ?

madphysicist:
Apologies. I have some code, but been out of the loop with work/kids for a bit. Will polish it up into PR material shortly.

MatthieuDartiailh:
Thanks @madphysicist. In the meantime I will merge this.

mcejp:
I am guessing the second pull request was never made?

madphysicist:
@mcejp You are right, it was never made. I have a bit more time nowadays, so I can still take a look, unless you would like someone else to take it over.

MatthieuDartiailh:
I implemented the opened resources idea some time ago https://github.com/pyvisa/pyvisa/pull/357

madphysicist:
@MatthieuDartiailh. Thank you for that. I just looked at the PR, and it looks very nice.

PR #281, opened 2017-10-12
Add support for different field / record separators

ap98nb26u:
Split functions do not work well when returning messages using different fields and record delimiters (e.g. ',', ';').  My customer's YOKOGAWA DL 9000 oscilloscope is working on this change.

### DIFF
--- a/pyvisa/util.py
+++ b/pyvisa/util.py
@@ -17,6 +17,7 @@ import functools
 import io
 import os
 import platform
+import re
 import sys
 import subprocess
 import warnings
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
You can treat such messages by passing a custom separator function to from_ascii_block. This use case seems pretty unusual to me and I am not sure it does justify to extend the acceptable values of separator. Furthermore your change prevent the use of multi-characters separators (even though they are likely unusual). 

If you still think this should be included in pyvisa, please update the function documentation and add tests.

MatthieuDartiailh:
Closing as I do feel this is truly needed (the custom separator function can handle this just fine)

PR #282, opened 2017-10-16
Slow data transfer for large data sets.

tomhb:
Change bytes to bytearray to improve efficiency in Python 3 when transferring large amounts of data.
Details of efficiency are in: http://ze.phyr.us/bytearray/

With some testing this issue may only exist in Python 3.6 and not in Python 2.7.


MatthieuDartiailh:
While I totally agree that this is a meaningful change, I am worried about breaking backwards compatibility by changing the return type of read_raw. I think it may be worth adding a _read_raw method returning a bytearray and simply wrap it to keep returning bytes from read_raw but use the unwrapped version in other functions. @hgrecco do you have an opinion here ?

MatthieuDartiailh:
Your last commit does not fix the issue with read_raw (still returning bytearray) and lead to a lot of duplication. I would prefer to have ```read_raw``` calling ```_read_raw``` and doing simply the conversion.

tomhb:
Thanks for the comments! I have updated read_raw to return bytes(_read_raw()). As per your previous comment _read_raw is called everywhere inside the class.

MatthieuDartiailh:
Thanks for your contribution @tomhb. Merging !

bors r+

PR #285, opened 2017-10-27
shell: support for 'termchar' command

skrchnavy:
Added support for termination character (read_termination) setup / read into shell (visa command line tool).

To display termination character
```
(open) termchar
```
To set termination character (read and write)
```
(open) termchar <termchar> [<termchar>]
```
`<termchar>` is one of `CR`, `LF`, `CRLF`, `NUL`.
To disable termination character, `None` shall be used
If two termchars are provided, first is read termination character and second is write termination character

### DIFF
--- a/docs/shell.rst
+++ b/docs/shell.rst
@@ -112,7 +112,9 @@ We can also get a list of all visa attributes::
 
 
 To simplify handling of VI_ATTR_TERMCHAR and VI_ATTR_TERMCHAR_EN, command 'termchar' can be used.
-To setup termchat to '\r' (CR or ascii code 10)::
+It also sets write termination character.
+
+To setup termchar to '\r' (CR or ascii code 10)::
 
     (open) termchar CR
     Done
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Here you only change the read_termchar and from experience both are useful (and sometimes the write even more than the read). SO If you want to add an easier access to those I would suggest adding both and also potentially supporting other characters if possible (I remember an instrument using '\0' as termchar).

skrchnavy:
My main intention was to simplify setting of termchar and termchar_en using one command, because termchar_en is False by default. (and also setting termchar for ni driver doesn't work for me, wrong type).
Setting write is possible, but I ma not sure if it shall be setup in common command or 2 different commands. If common set is OK as typical cas, I am OK with such change.
Other characters as termchar is possible via `attr` command, anyway if '\0' is used, it shall be added.

skrchnavy:
@MatthieuDartiailh 
I have updated implementation and also description of this PR with support for read and write termination. Also NUL is supported.

MatthieuDartiailh:
I made a minor modification in the docs but this looks good. Merging

bors r+

MatthieuDartiailh:
One minor thing I forgot before merging, to check that a key exist inside a dictionary the following syntax is more often used (compared to has_key):
```python
if key in mydict:
    pass
```

PR #286, opened 2017-10-27
visa executable console script

skrchnavy:
It is possible to install `visa` console command, so instead
```
python -m visa info
```
it is possible to call
```
visa info
```


MatthieuDartiailh:
Not sure how I feel about this one I fear it will confuse some users a bit more about the distinction between VISA and PyVISA. @hgrecco do you have an opinion ?

skrchnavy:
@MatthieuDartiailh I had same feeling, I was not sure how to name the command. It could be `pyvisa`, `pyvisacmd`. `pyvisashell`, ...
If you have better proposal I am OK with changing it.

MatthieuDartiailh:
Not sure. pyvisa-shell looks best to me but - is not allowed. pyvisa_shell would work I think. Can you try ?

MatthieuDartiailh:
Do you plan to complete this @skrchnavy ? Or should we close ?

MatthieuDartiailh:
Have you tested this @skrchnavy ? I remember reading that bash does not accept names containing  dashes, but I may be wrong.

skrchnavy:
I have updated the code ti 'install' these commands:
* `pyvisa-shell` as alias to `python -m visa shell`
* `pyvisa-info` as alias to `python -m visa info`

MatthieuDartiailh:
For future reference this may cause issues with some shells (https://unix.stackexchange.com/questions/168221/are-there-problems-with-hyphens-in-functions-aliases-and-executables) but as it is mostly an alias it should be fine. I will test and merge before the end of the week hopefully thanks again for your contributions.

skrchnavy:
I tested it on linux (ubuntu) in bash and there is no issue in hyphens.

I think most of current linux systems shall be OK with these names.

MatthieuDartiailh:
Testing on windows, I noted one issue in pyvisa-info. The installation process creates packages named pyvisa-shell-script and pyvisa-info-script that end up being listed as backends (because of their name). As a consequence we should filter the output of get `highlevel.list_backends` to remove them. It may be reasonable to remove any package ending by `-script` as it makes for an unrealistic name for an extension package I would say.

skrchnavy:
Good point of this side effect. For now I am not sure, how to improve backends check to ignore nonbackend modules.
Is there a point in backend that could be checked to ensure module is backend implementation?

MatthieuDartiailh:
Apart from attempting to get a get the library wrapper class I do not think there is. However relying on this means that in some cases it may not be obvious how to debug why a backend does not show. Honestly I think filtering based on the name should be fine (we can add a note in the documentation stating that extension must avoid names ending in `-script`.

skrchnavy:
I checked differences in result returned by `pkgutil.iter_modules()` and scripts are retuned as `ispkg=False` and backends as `ispkg=True`. Not sure if there is a way, how to implement backend as a module (not-package).
So the check could be:
```
if ispkg and name.startswith('pyvisa-')
```

Or do you prefer check by name?

MatthieuDartiailh:
I would prefer by name, I have one use case in mind in which a script could be used (case in which the implementation of the backend is "buried" in another package (one of my wip: i3py), and in which case the script is used simply to expose the extension to the pyvisa machinery).
We are dealing in corner cases here and so I prefer something for which I have no counter example.

skrchnavy:
OK, I will update check of backends using name.

skrchnavy:
check updated and tested on both windows and linux (python 2.7)

MatthieuDartiailh:
I added a note into the docs about the name filtering. If you are fine with it I will merge.

skrchnavy:
No comments.

MatthieuDartiailh:
Merging then. Thanks again.

bors r+

MatthieuDartiailh:
@skrchnavy thanks for your high quality work. If you have more time to dedicate to this project let me know, I can point some things that could use some work.

PR #290, opened 2017-11-07
Fixes in shell.py (reported in #285)

skrchnavy:
Fix based on comment in #285
> One minor thing I forgot before merging, to check that a key exist inside a dictionary the following syntax is more often used (compared to has_key): `if key in mydict: ...`

Also checking of read and write termination improved


MatthieuDartiailh:
Thanks for the cleanup !

bors r+

Issue #298, opened 2017-11-27
shell: incorrect handling of attribute types in set operation


skrchnavy:
Shell doesn't convert attribute value to proper type of attribute. :
* `attr VI_ATTR_TERMCHAR_EN` returns True/False as bool type when called after open
* `attr VI_ATTR_TERMCHAR_EN True` sets attribute as string/unicode type
* `attr VI_ATTR_TERMCHAR_EN` returns True/False as string / unicode type when called later.
There is no option to properly set type.

Proper type handling for attributes in shell shall be supported.

(tested with pyvisa-py as backend, ni-visa backend not tested)

skrchnavy:
#299 merged, closing as solved

PR #299, opened 2017-11-27
Shell: set attribute processing

skrchnavy:
converts value to python type based on visa attribute type.

Also return value from set_attribute is processed

Relates to #298


skrchnavy:
@MatthieuDartiailh please review. Not all types are converted, I checked attributes.py for types used in RW attributes.

MatthieuDartiailh:
This seems perfectly reasonable to me and is a nice improvement to the shell. I guess you tested this, so I am merging.

bors r+

PR #300, opened 2017-12-05
Drop support for EOL Python 2.6, 3.2 and 3.3

### LINKED ISSUE #295: Drop support for Python 2.6 and 3.2 and 3.3
Those versions are really old. @hgrecco do you have an opinion

hugovk:
Fixes #295.


hugovk:
```
branches:
  only:
```
This section in .travis.yml is an impediment to contributions because it means we can't test our feature branches before creating PRs, which might cause undue noise if a branch isn't yet ready for a PR.

MatthieuDartiailh:
ping @hgrecco 

In the absence of answer I will merge at the end of the month. @hugovk the restriction on the branches is their because otherwise teh temporary branches used by bors causes issue. I honestly do not mind an early PR as it is nice to know people are working on something. You can put [WIP] to let maintainers know that it is not ready and that's fine.

MatthieuDartiailh:
As nobody pronounced against it and since we are not enough maintainers to assume the burden of those old versions, I will merge. Thanks @hugovk 

bors r+

setup.py
```diff
@@ -32,9 +32,6 @@ def read(filename):
 if sys.version_info < (3, 4):
```
PeterPablo: shouldn't this also be removed, since support now starts at >= 3.4?

MatthieuDartiailh: Indeed thanks for pointing it out.

hugovk: What about Python 2.7?

MatthieuDartiailh: This one has not reached EOL and is still widely used. So I don't believe, we (pyvisa maintainers) will drop it before 2020-2021 or even a bit later. That said I won't fight to keep it longer than necessary as it increases the maintenance burden.

hugovk: Yes, still big numbers for 2.7 at https://github.com/pyvisa/pyvisa/issues/295#issuecomment-349346466. 

Python 2.7 doesn't have the `enum` module, but this `enum34` package backports it and is needed for 2.7.

It's probably not worth dropping 2.7 just to remove this dependency.

MatthieuDartiailh: I should have been more careful in checking that. I reverted my commit. Sorry @hugovk.

MatthieuDartiailh: I may edit the check later to avoid mentioning Python 3.4 to avoid future confusions.

hugovk: Good idea, a simple Py2 vs. Py3 check would be better!

PR #301, opened 2017-12-09
Message based improvements

### LINKED ISSUE #252: 'Can't concat bytes to str' error when calling write_ascii_values
It's very strange, as message.encode(enco) is bytes and util.to_ascii_block() returns string, so line 239 of messagebased.py `message = message.encode(enco) + block` fails with an error specified in the title. 
I use "d" for converter and "," for separator
Why is this happening, wasn't it tested at all or am I doing something wrong?

I could fix it trivially by calling .encode on the block variable.

MatthieuDartiailh:
~~This is WIP and untested so far~~ This has now been fully tested locally:
- implement HP binary header following @thliebig work ~~(I plan on adding tests)~~ done
- fix the issue in write_ascii_values by properly encoding the values in `utils.to_ascii_block` (closes #252)
- add noisy warnings to encourage migrations to new api for values reading/writing
- uniformize the API for reading values by providing read_ascii_values and read_binary_values

Also make all file concerned by the update be PEP8 compliant.

I hope to complete and test this before Christmas, but anybody is welcome to test and review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ python:
   - "3.6"
 
 install:
-  - pip install coverage
-  - pip install coveralls
+  - if [ $TRAVIS_PYTHON_VERSION == '2.7' ] || [ $TRAVIS_PYTHON_VERSION == '3.6' ]; then pip install numpy; fi
+  - pip install coverage coveralls
 
 script:
   - coverage run -p --source=pyvisa --omit="*test*","*compat*" setup.py test
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -44,10 +44,10 @@ master_doc = 'index'
 # General information about the project.
 project = 'PyVISA'
 author = 'PyVISA Authors'
+
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
-
 version = pkg_resources.get_distribution(project).version
 release = version
 this_year = datetime.date.today().year
--- a/docs/rvalues.rst
+++ b/docs/rvalues.rst
@@ -78,7 +78,7 @@ If you have doubles `d` in big endian the call will be::
 You can also specify the output container type, just as it was shown before.
 
 By default, PyVISA will assume that the data block is formatted according to
-the IEEE convention. If you instrument use HP data block you can pass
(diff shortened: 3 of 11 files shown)


MatthieuDartiailh:
The broken tests come from Python 2.6. As we are likely to drop support for it in January I will not try to fix it. This is ready for review from my end. ping @hgrecco, @skrchnavy

thliebig:
@MatthieuDartiailh, I have added a missing fixup for the improved header support:
https://github.com/thliebig/pyvisa/commit/f9879997ef87f6d903be63bb90ab9efa5d74a43e
It catches the case that a termination character happens to be the last valid data byte(s). It occurs only very very rarly, but it does and it then crashes the read... Please consider adding the path to this PR

MatthieuDartiailh:
Done. And I set you as author. This somehow always bothered me but as it worked in my tests as I forgot to spend more time on it. If you can do a complete review it would be awesome.

MatthieuDartiailh:
@thliebig Do you want to do a final review before I merge ?

MatthieuDartiailh:
Merging.

bors r+

rjollos:
`MessageBasedResource.values_format` is deprecated according to the warning issued when setting the attribute, but it is still documented in the section [Preconfiguring the transfer format](https://pyvisa.readthedocs.io/en/master/rvalues.html?highlight=values_format#preconfiguring-the-transfer-format).

Is there an alternate recommendation? Before discovering the attribute I was using a partial function to avoid specifying the attributes for every query:

```python
query = functools.partial(inst.query_ascii_values, separator=',', container=numpy.array)
```

MatthieuDartiailh:
This exact issue is being discussed in https://github.com/pyvisa/pyvisa/issues/327. The short answer is currently there is no way to preconfigure the parameters.

The docs definitively needs to be re-worked and indeed that mention should be removed. 

I am short of time currently so if you feel like to address any of these issues I will be happy to review PR.

rjollos:
> I am short of time currently so if you feel like to address any of these issues I will be happy to review PR.

Thanks for the reply. I would like to understand the pyvisa architecture and make some contributions, so I will try to prepare a pull request in the near future.

PR #307, opened 2018-01-18
Prefer import to pkgutil.iter_modules(), support PyInstaller

charlesnicholson:
`pyvisa.highlevel.get_wrapper_class` iterates through the discovered backends to find one that matches the requested name. This iteration is performed in `list_backends()` and uses `pkgutil.iter_modules()`. If a backend with the matching name is found, the code attempts to dynamically import it.

When PyInstaller "freezes" modules, they can't be found by `pkgutil` and friends, which means that even if you freeze pyvisa and pyvisa-py, the pyvisa-py backend can't be found because `iter_modules` can't see it.

This PR eliminates the search loop, which might not have been strictly necessary. It simply attempts to do the dynamic import. Success and failure are both handled identically to the previous version, but failure is observed when an ImportError is raised.

As a result, PyInstaller can now package up single executables containing pyvisa and pyvisa-py!

Note: `pyvisa.highlevel.list_backends()` still uses pkgutil, which will misreport inside of PyInstaller environments. PyInstaller has a long-term bug to provide their own Importer that knows how to work with pkgutil, but in the short term, this PR seems like an improvement.

### DIFF
--- a/pyvisa/highlevel.py
+++ b/pyvisa/highlevel.py
@@ -1445,14 +1445,6 @@ def get_wrapper_class(backend_name):
     except ImportError:
         raise ValueError('Wrapper not found: No package named pyvisa-%s' % backend_name)
 
-#    for pkgname in list_backends():
-#        if pkgname.endswith('-' + backend_name):
-#            pkg = __import__(pkgname)
-#            _WRAPPERS[backend_name] = cls = pkg.WRAPPER_CLASS
-#            return cls
-#    else:
-#        raise ValueError('Wrapper not found: No package named pyvisa-%s' % backend_name)
-
 
 def open_visa_library(specification):
     """Helper function to create a VISA library wrapper.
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
This looks quite reasonable to me indeed. I will merge in a week save if somebody comes up with a good argument against it.

charlesnicholson:
Hello! Just a friendly ping, it's been a week. Any thoughts on merging?

Also, do you have any thoughts on when you'll do another pypi release? We like to track pypi if possible; right now our internal system is just pointing at this branch on my fork... :)

MatthieuDartiailh:
Thanks for the ping. I will trigger bors.

@hgrecco gave me access to PyPI so I can now do releases. First I would like to get #300 and #301 in so if everything else goes smoothly I should be able to do a new release around mid-february and perhaps a bit before that.

bors r+

charlesnicholson:
Sounds great, thanks for the update.

martinbra:
As of today (1.10.0.dev), I had to force the pyinstaller to add the pyvisa-py module for the pyvisa to work:
```
>pyinstaller program.py -F 
>dist\program.exe
No package named pyvisa-py
```
```
>pyinstaller program.py -F --hidden-import pyvisa-py
>dist\program.exe
works fine
```

PR #308, opened 2018-01-19
Added message-length-reached check to _read_raw()

marcoep:
When using pyvisa-py as a backend and a TCPIP/Socket attached
instrument, the read_raw(N) function would throw error due to "timeout
reached", even though N or more bytes were read into the buffer.

Thus, to prevent the timeout error, _read_raw(N) does not call the
underlying read function anymore as soon as N or more bytes were
received.

### DIFF
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -314,6 +314,8 @@ class MessageBasedResource(Resource):
                                  self._resource_name, size, status)
                     chunk, status = self.visalib.read(self.session, size)
                     ret.extend(chunk)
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
The size argument to read_raw was never properly documented, but actually refers to chunk_size (number of bytes to read at a time from the instrument). Documenting it has been on my TODO list for some time. They have been several discussions about exposing a dedicated function to read a specified number of bytes. As I am currently working on the reading logic in #301 I will try to add it there. I will ping you when it is done so that you can give it a look.

PR #318, opened 2018-02-25
Ensure that __doc__ exists before adding to it.

### LINKED ISSUE #214: py2exe-compiled executable with optimizations fails
When optimizations are turned on with py2exe then PyVISA fails to import.  This same situation can be tested without using py2exe by turning on optimizations at the python command line:
`python -OO ./script.py`
Should you do this on a script that imports PyVISA then you'll see the error:

> TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'

This problem results from the docstrings (`__doc__` attribute) not being populated when optimizations are enabled.  There are several ways to solve this problem, but the easiest is likely to check for `__doc__` being None before attempting to add to the documentation.

compholio:
Fixes #214.

### DIFF
--- a/pyvisa/attributes.py
+++ b/pyvisa/attributes.py
@@ -41,7 +41,10 @@ class AttributeType(type):
         if not name.startswith('AttrVI_'):
             return
         cls.attribute_id = getattr(constants, cls.visa_name)
-        if cls.__doc__: cls.redoc()
+        # Check that the docstring are populated before extending them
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
I added some more comments. This looks good bors will merge.

bors r+

MatthieuDartiailh:
Thanks @compholio

Issue #280, opened 2017-10-04
Regular expression syntax is not documented


awneil:
The Resource Manager documentation for list_resources() and list_resources_info() states that the 'query' parameter is a regular expression - but gives no definition or description or reference for the syntax of that regular expression

http://pyvisa.readthedocs.io/en/stable/api/resourcemanager.html#pyvisa.highlevel.ResourceManager.list_resources

I did eventually stumble upon this: https://github.com/pyvisa/pyvisa/blob/master/pyvisa/rname.py - which does contain a definition of the regular expression syntax.

It also points out that the regular expression syntax used here is **_not_** the same as the standard Python regular expression syntax - so it is all the more important that this should be stated in the documentation!

awneil:
The VISA Resource Regular Expression is defined (with examples) in **VPP-4.3: The VISA Library specification** 
Section 4.4.1
Tables 4.4.1 - 4.4.4

http://www.ivifoundation.org/specifications/


See also:  http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/findingvisaresourcesusingregularexpressions/

And: http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/visaresourcesyntaxandexamples/

MatthieuDartiailh:
Would you mind doing a pull request to add this to the docs ?

awneil:
I wouldn't mind, but someone would have to walk me through that process - I don't do Git

MatthieuDartiailh:
First you need to fork the repo (Fork button on the Github UI), then clone locally the forked repository (that you own). Create a new branch (regular-docs could be a good name but it does not really matter). Update the documentation files that need to be updated. Then commit your changes and push them to your repo. Once this is done the Github UI on your repo will ask you if you want to open a PR, just click on the button. If you are looking for git graphical interface, I would recommend Smartgit.

awneil:
OK ...

MatthieuDartiailh:
If you have more questions do not hesitate. Actually another simpler option would be to fork the project then simply edit the docs directly on Github, once you are done Github will handle the commit and branch creation for you and you will just have to create the PR.
If you can't do it I will probably do it at some point but I cannot say when.

awneil:
Is it sufficient just to update the docstrings for list_resources() and list_resources_info() in pyvisa/highlevel.py ?

MatthieuDartiailh:
[...]

PR #320, opened 2018-03-15
#280: Add VISA Resource Regular Expression syntax description in list…

awneil:
…_resources(),

and list_resources_info()

### DIFF
--- a/pyvisa/ctwrapper/highlevel.py
+++ b/pyvisa/ctwrapper/highlevel.py
@@ -196,7 +196,45 @@ class NIVisaLibrary(highlevel.VisaLibraryBase):
     def list_resources(self, session, query='?*::INSTR'):
         """Returns a tuple of all connected devices matching query.
 
-        :param query: regular expression used to match devices.
+        note: The query uses the VISA Resource Regular Expression syntax - which is not the same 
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
This looks great ! Thanks for putting it together !

Could also add a quick note (not the full explanation) to https://github.com/pyvisa/pyvisa/blob/master/docs/tutorial.rst ? In particular insisting on the fact that the default query does not return anything else than INSTR (so no RAW or SOCKET which is a usual question). If you can't, let me know and I will do it once I merged this.

awneil:
Sorry it took so long!

Could you point me to a reference for the formatting in a .rst file ?

MatthieuDartiailh:
I found some time to work on the docs. I will merge as is and open a PR for the docs update after. Thanks for your time @awneil.

bors r+

awneil:
ah - I hadn't added the note in the .RST.

So should I now make a new branch to do that?

MatthieuDartiailh:
Do not worry I added the note in #323 which is waiting for a small test on my end before it gets merged.

awneil:
ok - thanks.

:+1:

PR #336, opened 2018-05-15
rname: fix default gpib secondary address

MatthieuDartiailh:
The documented value in the header is \xFFFF but used in resource string it does crash. The Matworks link suggest using 0 in the absence of secondary address which works (tested on one instrument) and seems more logical.


MatthieuDartiailh:
In the absence of any complaints, I will merge since this fix a well identified issue and I was able to test the fix locally.

bors r+

Issue #338, opened 2018-05-20
PyVISA exceptions are not hashable


d-beckmann:
PyVISA exceptions override `__eq__`, but not `__hash__`, which makes them unhashable. This leads to problems with traceback printing (see https://bugs.python.org/issue28603). While the traceback module has been fixed in Python 3.6.4, this may still cause problems for users of older Python versions, or elsewhere. As far as I can see, `__eq__` has been introduced only for testing pickle/unpickle, and could be easily moved to the unittest (see attached patch).

https://gist.github.com/d-beckmann/2c9b6651d5b5b46b5e1af8946c855622

PR #340, opened 2018-05-27
pyvisa: make exception hashable by removing exception __eq__

MatthieuDartiailh:
I am not yet sure that this is without consequences, but I suppose it is. Any opinion ?


MatthieuDartiailh:
Since I believe this is the right thing to do and in the absence of comments, I am merging this.

Closes #338 

bors r+

PR #341, opened 2018-05-31
For default backend, use pyvisa-py if the NI binary cannot be found.

### LINKED ISSUE #334: pyvisa-py as fallback implementation
To use pyvisa, one needs a backend, by default the NI-VISA library. Unfortunately, the binary nowadays seems to be behind a [login wall](http://lumen.ni.com/nicif/US/GB_NIDU/content.xhtml?du=http://www.ni.com/download/ni-visa-5.4.1/4626/en/), so there is definitely a barrier to get going.

If no NI-VISA library is found installed, I think it would make sense to automatically fall back to the open [pyvisa-py](https://github.com/pyvisa/pyvisa-py) implementation. This way, nothing changes for users having NI-VISA installed, and it's still used by default if people can install the library, but the procedure is smoother than now for people who don't have access to that library.

In particular, it would make it easier to use packages like pymeasure where you are [not necessarily using](http://pymeasure.readthedocs.io/en/latest/tutorial/connecting.html) a `ResourceManager` to configure the backend choice. Yes I know that it's possible to select the backend via environment variable, but that's another extra step that users have to do every time they use pyvisa.

Does the proposal to automatically fall back to pyvisa-py if NI-VISA is not installed sound good/smart/acceptable?

bilderbuchi:
This fixes #334.

I also have a commit prepared that adds `pyvisa-py` as a dependency to setup.py, but I don't know if you want that. If we include it, it's guaranteed that people will always have a backend available even if the NI one isn't installed, so i lean towards doing that. Please comment on that @MatthieuDartiailh (*edit: I have pushed that commit to the PR now, it should be easy to not included it if desired*)

I tested this on my local machine (without the NI library). pyvisa-py gets selected successfully:

```
import pyvisa
pyvisa.ResourceManager()
Out[2]: <ResourceManager(<VisaLibrary('unset')>)>
```

If neither NI nor pyvisa-py are around, we get (with debug output enabled)
```
pyvisa.ResourceManager()
DEBUG:pyvisa:No visa library specified, trying to find alternatives.
DEBUG:pyvisa:Environment variable PYVISA_LIBRARY is unset.
DEBUG:pyvisa:No user defined library files
DEBUG:pyvisa:Automatically found library files: []
DEBUG:pyvisa:Did not find NI binary
DEBUG:pyvisa:Did not find pyvisa-py package
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<snip>
ValueError: Could not locate a VISA implementation. Install either the NI binary or pyvisa-py.
```

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -3,15 +3,15 @@
 Configuring the backend
 ============================
 
-Currently there are two backends available: The one included in pyvisa which
-uses the NI library. This is used by default and the configuration is described
-in the next chapter.
-And then there is pyvisa-py a pure python implementation of the VISA libary.
-It can be selected by passing a parameter to the ResourceManager:
+Currently there are two backends available: The one included in pyvisa, which
+uses the NI library, and the backend provided by pyvisa-py, which is a pure python implementation of the VISA library.
+If no backend is specified, pyvisa uses the NI backend if the NI library has been installed (see next section for details). Failing that, it uses the pyvisa-py backend.
+
+You can also select a desired backend by passing a parameter to the ResourceManager, shown here for pyvisa-py:
 
     >>> visa.ResourceManager('@py')
 
-Alternativly it can also be selected by setting the environment variable
+Alternatively it can also be selected by setting the environment variable
 PYVISA_LIBRARY. It takes the same values as the ResourceManager constructor.
 
 Configuring the NI backend
--- a/pyvisa/highlevel.py
+++ b/pyvisa/highlevel.py
@@ -1446,6 +1446,30 @@ def get_wrapper_class(backend_name):
         raise ValueError('Wrapper not found: No package named pyvisa-%s' % backend_name)
 
 
+def _get_default_wrapper():
+    """Return an available default VISA wrapper as a string ('ni' or 'py').
+
+    Use NI if the binary is found, else try to use pyvisa-py.
+    If neither can be found, raise a ValueError.
+    """
+
+    from .ctwrapper import NIVisaLibrary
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
I like this patch a lot. However I am not a big fan of having a kind of circular import between pyvisa and pyvisa-py. Furthermore your error message is very clear and pyvisa-py is not really stable so I would prefer to revert your second commit.

bilderbuchi:
I searched online about circular dependencies (it worried me too) but discussions indicated that it is fine as long as you don't have impossible constraints. Also, as it is green, the CI seems to have not found a problem when installing, right?
Anyway, I will remove that commit.

MatthieuDartiailh:
Thanks. I will trigger bors as soon as you revert.

bilderbuchi:
OK, I just force-pushed to remove the commit.

MatthieuDartiailh:
Perfect !

bors r+

bilderbuchi:
Thank you!

> pyvisa-py is not really stable

Speaking of which, are you going to release another version (soon)? 0.2 seems to be from 2015, but I see activity in the commits list. Is it workable to use in 0.2?

MatthieuDartiailh:
No 0.2 is not really workable. But I lack time to test https://github.com/pyvisa/pyvisa-py/pull/139 that I would like to include in the release if you some spare time and the right instrument it would be great if you could test. Once merged I will try to release quickly.

bilderbuchi:
Thanks. Sorry, don't have a fitting device for this.

MatthieuDartiailh:
Too bad

PR #363, opened 2018-09-28
Added possibility to redirect logging to file

efferre79:
Without breaking compatibility with older code I have kept the log_to_screen() function and added a more generic log() function to be able to redirect the specific pyvisa logging output to file (in the case an application is already using stderr for other purposes).

### DIFF
--- a/pyvisa/__init__.py
+++ b/pyvisa/__init__.py
@@ -23,8 +23,11 @@ logger.addHandler(compat.NullHandler())
 
 
 def log_to_screen(level=logging.DEBUG):
+    log(None, level) # sys.stderr by default
+
+def log(stream_output, level=logging.DEBUG):
     logger.setLevel(level)
-    ch = logging.StreamHandler()
+    ch = logging.StreamHandler(stream_output)
     ch.setLevel(level)
 
     formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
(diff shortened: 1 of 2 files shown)


pyvisa/__init__.py
```diff
@@ -23,8 +23,11 @@
 
 
 def log_to_screen(level=logging.DEBUG):
+    log(None, level) # sys.stderr by default
+
+def log(stream_output, level=logging.DEBUG):
```
MatthieuDartiailh: To avoid any confusion on what this function do I would prefer for it to be called `log_to_stream`. Otherwise people may think that it actually log a message.

MatthieuDartiailh:
Could you add an entry to the changelog ? Doing this as part of the PR make the release process easier when the time comes. Once this is done I will merge.

efferre79:
> Could you add an entry to the changelog ? Doing this as part of the PR make the release process easier when the time comes. Once this is done I will merge.

Should I amend the latest commit message or this conversation?

MatthieuDartiailh:
No you should simply add a line in the file CHANGES under 1.10 with a reference to this PR.

MatthieuDartiailh:
Thanks for your responsiveness !!!

MatthieuDartiailh:
Thanks

bors r+

MatthieuDartiailh:
Trying again....

bors r+

MatthieuDartiailh:
Bors was blocked for no apparent reason so I bypassed it fro once (Travis was green on the test branch)

Issue #370, opened 2018-10-11
Possible pyvisa.constants duplicates 


LaurentRDC:
Right now, in the `constants` module, the following enumerations are defined:
```
class StatusCode(enum.IntEnum):
    (…omitted...)
    #: Specified offset is not accessible from this hardware.
    error_nonsupported_offset = VI_ERROR_NSUP_OFFSET

    #: The specified offset is not properly aligned for the access width of the operation.
    error_nonsupported_offset_alignment = VI_ERROR_NSUP_OFFSET
    (…omitted...)
```

Is there a reason that `VI_ERROR_NSUP_ALIGN_OFFSET` is not used above for `error_nonsupported_offset_alignment`? It is defined in `pyvisa.constants` but not used.

If this is indeed a typo, I suggest that the enumerations in `pyvisa.constants` make use of the [`enum.unique`](https://docs.python.org/3/library/enum.html#enum.unique) decorator to ensure this doesn't happen by mistake.

LaurentRDC:
There is nothing in the [specifications](http://www.ivifoundation.org/docs/vpp432.pdf) that indicate these two enums should map to the same value.

MatthieuDartiailh:
This does look like a typo. Could you make a PR (with enum.unique) ? If you do please add a mention on the changelog.

MatthieuDartiailh:
Closed by #371

PR #371, opened 2018-10-11
Fixed duplicate StatusCode value and made all enums unique

LaurentRDC:
Fix for issue #370 .

The enums `StatusCode.error_nonsupported_offset` and `StatusCode.error_nonsupported_offset_alignment` had the same value of `VI_ERROR_NSUP_OFFSET`, and `VI_ERROR_NSUP_ALIGN_OFFSET` was not used.

To ensure this does not happen again,, all enum classes in the `pyvisa.constants` modules have been decorated with the [`enum.unique`](https://docs.python.org/3.5/library/enum.html?highlight=enum#enum.unique) class decorator.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ Other contributors, listed alphabetically, are:
 * Gregor Thalhammer <gregor.thalhammer@gmail.com>
 * HummingBrid
 * kcsaff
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
Great ! Thanks for taking the time to do this.

bors r+

PR #373, opened 2018-10-20
Use ni backend when specifying a file in open_visa_library

lupien:
The changes (commit f2d69892) in open_visa_library to use either a NI backend or if not available the pyvisa-py are nice except they seem a bit too broad.

If somebody specifies a file (dll) to use when opening the ResourceManager, it still looks for the standard dll to see if they are available (_get_default_wrapper()). Therefore, if none of the standard dll are available (like visa32.dll, visa64.dll on windows) but a correct other dll is given (like agvisa32.dll for the agilent visa library) it still fails.
It is probably a rare condition to be in but it is not impossible.

This PR fixes that.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ PyVISA Changelog
 - added new function log_to_stream() PR #363
 - Made all enumerations of the `constants` module unique. 
   Fixed duplicate enums in StatusCode PR #371
+- Use ni backend when specifying a file in open_visa_library PR #373
 
 
 1.9.1 (2018-08-13)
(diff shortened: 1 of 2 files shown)


lupien:
I updated the CHANGES file

MatthieuDartiailh:
Perfect

bors r+

MatthieuDartiailh:
I do not know what is going on here Travis did not pick up your last commit and bors is not answering. I will merge directly.

MatthieuDartiailh:
Ok bors was only slow ...

PR #382, opened 2018-11-27
ctwrapper: use context manager to open (and close) pipes

mgubser:
If the pipes are left open when trying to find the *visa* library paths, we get a lot of exceptions on Linux similar to this:

```
/usr/local/lib/python3.6/subprocess.py:766: ResourceWarning: subprocess 2020 is still running
  ResourceWarning, source=self)
/home/.../.venv/lib/python3.6/site-packages/pyvisa/ctwrapper/cthelper.py:68: ResourceWarning: unclosed file <_io.TextIOWrapper name=34 encoding='UTF-8'>
  _os.popen('/sbin/ldconfig -p 2>/dev/null').read())
/usr/local/lib/python3.6/subprocess.py:766: ResourceWarning: subprocess 2022 is still running
  ResourceWarning, source=self)
/home/.../.venv/lib/python3.6/site-packages/pyvisa/ctwrapper/cthelper.py:74: ResourceWarning: unclosed file <_io.TextIOWrapper name=34 encoding='UTF-8'>
  res = re.search(expr, _os.popen(cmd).read())
```

By using the provided context manager this can be prevented.

### DIFF
--- a/pyvisa/ctwrapper/cthelper.py
+++ b/pyvisa/ctwrapper/cthelper.py
@@ -62,12 +62,13 @@ if _os.name == "posix" and _sys.platform.startswith('linux'):
         def _findlib_ldconfig(name):
             # XXX assuming GLIBC's ldconfig (with option -p)
             expr = r'/[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
-            res = re.search(expr,
-                            _os.popen('/sbin/ldconfig -p 2>/dev/null').read())
+            with _os.popen('/sbin/ldconfig -p 2>/dev/null') as pipe:
+                res = re.search(expr, pipe.read())
             if not res:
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Why did you close and delete ? I had not merged !

mgubser:
It was an exact duplicate of #380, which you already approved. Sorry for not checking first.

mgubser:
I verified that the changes in ``pyvisa`` work with Python 3.6 and that they produce the desired effect (i.e. not printing the error messages by closing the pipes appropriately). I then checked using a Python console that ``popen`` works with a context manager in 2.7. So I guess that will work in ``pyvisa`` as well.

MatthieuDartiailh:
Sounds good. Thanks for checking @mgubser.

bors r+

PR #384, opened 2018-12-01
Add attr command omitted from example

rjollos:


MatthieuDartiailh:
This looks like a pertinent correction. Merging.

bors r+

Issue #398, opened 2019-03-05
visa.ResourceManager('@ni') returns either visa64.dll or visa32.dll, choosing randomly


HansJA:
System info:
Python 3.7.2
PyVISA 1.9.1
NI-VISA full 64bit install v18.5
W10 fully updated as of this morning

Problem: visa.ResourceManager('@ni') returns either visa64.dll or visa32.dll, choosing randomly.

one attempt:
C:\Users\Hans\Desktop>python
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import visa
>>> visa.ResourceManager('@ni')
<ResourceManager(<VisaLibrary('C:\\WINDOWS\\system32\\visa64.dll')>)>
>>> ^Z

next attempt
C:\Users\Hans\Desktop>python
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import visa
>>> visa.ResourceManager('@ni')
<ResourceManager(<VisaLibrary('C:\\WINDOWS\\system32\\visa32.dll')>)>
>>> ^Z

In 10 attempts visa32 was chosen 6 times, visa64 4 times.

visa32 gets chosen more often, my feeling is that in the longer perspective visa32.dll is chosen 7 or 8 times out of 10.

MatthieuDartiailh:
Do you see any resulting issue ? According to the following thread it is possible for both to actually be 64 bits and should not cause any trouble in your application? https://forums.ni.com/t5/Instrument-Control-GPIB-Serial/NI-MAX-requires-32-bit-VISA-missing-in-64-bit-installs/m-p/3896195

You can confirm the bitness by running multiple times `pyvisa-info` from the command line.

HansJA:
Is bitness your only criterion for selecting a resource manager?
I think you should assume that NI chose different names for good reason.

Other than that, I have multiple issues with PyVISA, but I cannot safely
attribute any one problem with this bug report.
Most if my issues have one specific form:
I do
pip install -U <all pyvisa related packages>
python
import visa
<Get an instrument and try to use it>
^Z
pip uninstall <all pyvisa related packages>

pip install - U <all pyvisa related packages>
python
[...]

PR #399, opened 2019-03-06
Reworked search for library path to make it deterministic

hgrecco:
See #398

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,11 +4,12 @@ PyVISA Changelog
 1.10 (unreleased)
 -----------------
 
+- make the ordering of the visa library deterministic PR #399
 - properly close pipes when looking for a shared libary path on linux #380
 - fixing missing argument for USBInstrument.usb_control_out PR #353
 - usb_control_out -> control_out. warnings for deprecated usb_control_out PR #353
 - added new function log_to_stream() PR #363
-- Made all enumerations of the `constants` module unique. 
+- Made all enumerations of the `constants` module unique.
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Thanks @hgrecco. Would it also make change to change the order so that we select 64 on 64 architecture ? I think it would reduce surprise from our users but I don't know if it could cause issues.

hgrecco:
I think we should investigate that road but I would do it in a different PR to make it easier to revert. 

In particular, my suggestion would be to reorder the libraries based on bitness (instead of filtering). In this way,  the most appropriate library will be used when [looping through the options](https://github.com/pyvisa/pyvisa/blob/master/pyvisa/highlevel.py#L88). Filtering the list has failed in the past (although it is true that the ecosystem was more balkanized then)

As I mentioned in #398, opening the library is still the best way to test that it works. I think that additionally we could put [these checks](https://github.com/pyvisa/pyvisa/blob/master/pyvisa/ctwrapper/highlevel.py#L110) in the reordering.

MatthieuDartiailh:
If we simply order based on bitness we can still get visa32 which can be confusing. But yes we can investigate in a different PR.

MatthieuDartiailh:
@hgrecco could you update the releasenotes with that change ? I am trying to keep them up to dates at all time to avoid the pain to do everything at release time.

MatthieuDartiailh:
I took care of the changelog I will merge before the end of the week.

Issue #414, opened 2019-05-28
List of open Resources


ruggiamp:
Hi,

I would like to know how to retrieve a list of just the resources that are already open.
Or, maybe, how do I ask a resource if it is open or not.

Many thanks
Ruggero

MatthieuDartiailh:
There is no way to do that in 1.9.1. If you are willing to work on the master branch you can access the open resources from the resource manager (attribute `_created_resources`).

Could you describe your use case ? I am wondering if we should expose a Public API to do that.

This feature is expected to be released in the next release (in July a priori)

ruggiamp:
Thanks for the promt answer!
I am dealing with many devices at the same time and I might open and close them in different moments.
I just wanted to have the possibility to know in realtime what is the situation: how many and which devices are open. 
I ended up dealing with that by locally keeping track of each resource. It works but it would be better if there was a method already implemented in the ResourceManager I think.

Have a nice day!
Ruggero

MatthieuDartiailh:
I could add a public method returning a snapshot of the resources opened at the time of the call. `list_opened_resources` for example returning a list based on the WeakSet the resource manager keeps for book keeping purposes. WDYT ?

ruggiamp:
It seems good!
I think it would be great if the output was just a list of the open devices, without any additional information, so to be easily readable...

MatthieuDartiailh:
Do you mean a list of resource names ? I feel it would be more flexible to return the actual resource objects. Then people can do `[r.resource_name for r in rm.list_opened_resources]`. (I need to check if `resource_name` is publicly accessible).

ruggiamp:
Yep, I meant a list of resource names. I didn't know that possibility...

MatthieuDartiailh:
Addressed by #415

PR #415, opened 2019-05-31
ResourceManager.list_opened_resources

MatthieuDartiailh:
Based on @ruggiamp suggestion in #414, add a `list_opened_resources` method to the resource manager returning a snapshot of the opened resources. The documentation has been updated to add an example of its usage.

Also in the Resource class we use consistently the resource name stored in the private attribute (`_resource_name`) to avoid relying on the VisaLibrary object to get it. This should fix issues such as https://github.com/pyvisa/pyvisa-py/issues/151.

@ruggiamp please test if you can (I already tested locally).

@hgrecco @tivek please review if you can.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ PyVISA Changelog
 1.10 (unreleased)
 -----------------
 
+- add a list_opened_resources method to the ResourceManager PR #415
+- use privately stored resource name in Resource class rather than relying on
+  VisaLibrary PR #415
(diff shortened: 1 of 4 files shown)


ruggiamp:
Many thanks! 
I have seen the example in the documentation. Good job!

Best

MatthieuDartiailh:
Thanks. If you have time try to test the release candidate.

PR #429, opened 2019-06-28
Fix partial reads in binary data which cause VI_ERROR_CONN_LOST when …

bjaraujo:
…using HiSLIP protocol and large data sizes.


bjaraujo:
Forgot to add case when there is no header.

PR #430, opened 2019-06-28
HISLIP - Fix partial reads in binary data which cause VI_ERROR_CONN_LOST

bjaraujo:
Fix partial reads in binary data which cause VI_ERROR_CONN_LOST when using HiSLIP protocol and large data sizes. Added case for when there is no header to read data as before.

### DIFF
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -495,7 +495,7 @@ class MessageBasedResource(Resource):
         if data_length is not None:
             block.extend(self.read_bytes(data_length))
         else:
-            block.extend(self.read_raw())
+            block.extend(self._read_raw())
 
         if expect_termination and self._read_termination is not None:
             expected_length += len(self._read_termination)
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -503,7 +503,7 @@ class MessageBasedResource(Resource):
 
             expected_length = offset + data_length
 
-        else header_fmt == 'empty':
+        elif header_fmt == 'empty':
 
             block = block + self._read_raw()
 
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -503,6 +503,10 @@ class MessageBasedResource(Resource):
 
             expected_length = offset + data_length
 
+        else header_fmt == 'empty':
+
+            block = block + self._read_raw()
+
         if expect_termination and self._read_termination is not None:
             expected_length += len(self._read_termination)
 
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -469,43 +469,33 @@ class MessageBasedResource(Resource):
         :rtype: type(container)
 
         """
-        block = b''
+        block = bytearray()
         expected_length = 0
 
         if header_fmt == 'ieee':
-            hash = self.read_bytes(1)                   # Hash '#'
-            digits = self.read_bytes(1)                 # Digits
-            block_size = self.read_bytes(int(digits))   # Block size
+            header = self.read_bytes(2)                    # Hash '#' and digits
+            digits = int(header[1:])                       # Digits
+            block_size = self.read_bytes(digits)           # Block size
+            header += block_size
+            block.extend(header)
 
-            block = block + hash
-            block = block + digits
-            block = block + block_size
-
-            block = block + self.read_bytes(int(block_size))
-
-            offset = 2 + int(digits)
-            data_length = int(block_size)
+            offset, data_length = util.parse_ieee_block_header(header)
 
             expected_length = offset + data_length
 
         elif header_fmt == 'hp':
 
-            hash = self.read_bytes(2)                   # Hash '#A'
-            block_size = self.read_bytes(4)             # Block size
+            header = self.read_bytes(6)                     # Hash '#A' + block size
+            block.extend(header)
 
-            block = block + hash
-            block = block + block_size
-
-            offset = 6
-            data_length = int_from_bytes(block_size,
-                                 byteorder='big' if is_big_endian else 'little'
-                                 )
+            offset, data_length = util.parse_hp_block_header(header, is_big_endian)
 
             expected_length = offset + data_length
 
-        elif header_fmt == 'empty':
-
-            block = block + self._read_raw()
+        if data_length is not None:
+            block.extend(self.read_bytes(data_length))
+        else:
+            block.extend(self.read_raw())
 
         if expect_termination and self._read_termination is not None:
             expected_length += len(self._read_termination)
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -469,15 +469,38 @@ class MessageBasedResource(Resource):
         :rtype: type(container)
 
         """
-        block = self._read_raw()
+        block = b''
         expected_length = 0
 
         if header_fmt == 'ieee':
-            offset, data_length = util.parse_ieee_block_header(block)
+            hash = self.read_bytes(1)                   # Hash '#'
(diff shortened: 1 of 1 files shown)


pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
 
         """
-        block = self._read_raw()
+        block = b''
         expected_length = 0
 
         if header_fmt == 'ieee':
-            offset, data_length = util.parse_ieee_block_header(block)
+            hash = self.read_bytes(1)                   # Hash '#'
+            digits = self.read_bytes(1)                 # Digits
```
MatthieuDartiailh: Reading multiple times is not free, we should group those readings as much as possible. This was actually the motivation behind the current code: avoid multiple reading when possible by reading a large chunk. However it this causes issues I am fine changing it but we should try to limit the overhead.

pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
+            block_size = self.read_bytes(int(digits))   # Block size
+
+            block = block + hash
+            block = block + digits
+            block = block + block_size
+
+            block = block + self.read_bytes(int(block_size))
```
MatthieuDartiailh: We should not read all the data at this stage. It is too early in particular because we did not yet address the question of whether or not the termination character should be read.

pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
         """
-        block = self._read_raw()
+        block = b''
         expected_length = 0
 
         if header_fmt == 'ieee':
-            offset, data_length = util.parse_ieee_block_header(block)
+            hash = self.read_bytes(1)                   # Hash '#'
+            digits = self.read_bytes(1)                 # Digits
+            block_size = self.read_bytes(int(digits))   # Block size
```
MatthieuDartiailh: We need to test for digits being 0 here. The Keithley 2000 for example return such a block which is not technically ieee compliant but is close enough. To allow that kind of use case, we could add an `expected_length` keyword argument to `read_binary_values` and `query_binary_values`, that would default to `None`, allowing to write: `expected_length = expected_length or int(block_size)`.

pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
+            data_length = int(block_size)
+
             expected_length = offset + data_length
+
         elif header_fmt == 'hp':
-            offset, data_length = util.parse_hp_block_header(block,
-                                                             is_big_endian)
+
+            hash = self.read_bytes(2)                   # Hash '#A'
+            block_size = self.read_bytes(4)             # Block size
```
MatthieuDartiailh: Once again it makes sense to read all 6 bytes at once and then simply use the `parse_hp_block_header` utility to avoid code duplication.

pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
+            block = block + block_size
+
+            offset = 6
+            data_length = int_from_bytes(block_size,
+                                 byteorder='big' if is_big_endian else 'little'
+                                 )
+
             expected_length = offset + data_length
 
+        elif header_fmt == 'empty':
```
MatthieuDartiailh: This can be removed, see below.

pyvisa/resources/messagebased.py
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
...
+
+            block = block + self._read_raw()
+
         if expect_termination and self._read_termination is not None:
             expected_length += len(self._read_termination)
```
MatthieuDartiailh: The following should be replaced by something like:
```python
if expected_length is not None:
    block.extend(self.read_bytes(expected_length)
else:
    block.extend(self.read_raw())

File: pyvisa/resources/messagebased.py [hunk 7]
```diff
@@ -469,17 +469,44 @@ def read_binary_values(self, datatype='f', is_big_endian=False,
         :rtype: type(container)
 
         """
-        block = self._read_raw()
+        block = b''
```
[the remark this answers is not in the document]
bjaraujo: I can change that to a bytearray.

pyvisa/resources/messagebased.py
bjaraujo: I can combine those 2 into 1 call

pyvisa/resources/messagebased.py
bjaraujo: This comment I do not understand. The previous code would read too early. This one only reads the amount of data it "knows" that should be there. Isn't the termination after that?

MatthieuDartiailh: My point is that no matter the protocol at one point we need to read a block of data. It make more sense to go through the same code for all protocols to actually read the data once we determined the proper size on a protocol per protocol basis. See the modification I suggest to the reading below.

pyvisa/resources/messagebased.py
bjaraujo: The case for 0 has been tested. If it is zero it will read 0 bytes so remain the same.

bjaraujo:
So I have made the changes requested. It uses bytearray, less reads and uses utils to parse the header. I think this is better than passing chunk_size as parameter because I might not know the data size I am about to get.

MatthieuDartiailh:
I added a couple of options I mentioned earlier. We sadly do not have unit tests so the more thorough tests you can give this the better.

Coming back to `chunk_size`, if you read the code for `read_bytes` (https://github.com/pyvisa/pyvisa/blob/master/pyvisa/resources/messagebased.py#L342) you will see that we do not read everything at once, so controlling the chunk size may help speed the transfer. Also this is why I am wondering if there is not something more to this issue, since we always read in multiple time but somehow if we read a large block till the term char at first it causes an error.

bjaraujo:
I agree there is something strange and the failure is inside VISA so it is hard to understand what is wrong. Unfortunately there are not many ways of testing binary data but I will test it further.

MatthieuDartiailh:
The issue is that unit tests r equire

MatthieuDartiailh:
If you can try experimenting with the chunk size, it may help us understand what is going on. I will also try to test on a couple of instruments.

bjaraujo:
I think the patch works better than the original code for most cases but still not sure. This would need more testing. I think the problem might be due to VISA bug and by fixing it in pyvisa might be only a workaround.

MatthieuDartiailh:
Did you get to experiment with using various chunk size ? Do you see any speed difference ? I will run some tests later today (including tests on a Keithley 2000 that uses the new data size). Could you give me the model of the instrument you use for testing. It may be worth discussing with the manufacturer ?

Out of curiosity could you reproduce the same error using NI-MAX if you read the  same chunk size as PyVISA does currently ?

bjaraujo:
I am closing this issue. I think further investigation is required. This may be VISA related rather than pyvisa. I will keep my branch as experimental.

MatthieuDartiailh:
Most of the changes proposed in this PR make sense anyway (the only negative point is the multiple reads), so I will try to get something along those line for the next release. Please let me know of what you manage to dig. Also if your instrument happens to be from Keysight I have some contacts there that may help us investigate, so let me know.

bjaraujo:
Just to get back on this. I think chunk_size is determining factor here. Just changing the value of chunk_size solves the issues for most cases. So passing it in as option is probably the best idea.

MatthieuDartiailh:
Ok Thanks for the update. I will try to put together a PR soon. I you can test it, it will be great.

PR #498, opened 2020-03-03
Fix `delay` parameter for the query* methods.

daytonb:
There is a note that mistakenly called the `delay` parameter of the `query*` methods `query`.


MatthieuDartiailh:
Thanks !

bors r+

PR #500, opened 2020-03-10
Moved the metadata into setup.cfg

KOLANICH:


MatthieuDartiailh:
Thanks for the PR. Could you motivate a bit the change ? the toml is known to break the use of development mode through pip and looking at the changes I am not sure what we gain from this.

KOLANICH:
>the toml is known to break the use of development mode

I haven't tested this lib for it, but for my other libs using setup.cfg + pyproject.toml `pip3 install --upgrade -e .` works fine.

>Thanks for the PR. Could you motivate a bit the change ? 

`setup.cfg` is a declarative format that can be parsed. `setuptools_scm` fetches versions from git tags automatically and also writes commits hashes into non-release vers, which is convenient.

MatthieuDartiailh:
Then I will check. Do you know what version of pip you are using ? I was under the impression that using development mode required passing the --no-pep-517 flag on the most recent pip versions.

KOLANICH:
>I was under the impression that using development mode required passing the --no-pep-517 flag on the most recent pip versions.

This may happen if you disabled fetching packages from pypi. If it is the case, one needs also disable build isolation (imho disabling build isolation is a good idea anyway).

setup.cfg
```diff
@@ -1,3 +1,66 @@
...
+test_suite = pyvisa.testsuite.testsuite
+setup_require = setuptools>=42; wheel; setuptools_scm[toml]>=3.4.3
+
+[options.entry_points]
+console_scripts =
+    pyvisa-shell=pyvisa.cmd_line_tools:visa_shell
+    pyvisa-info=pyvisa.cmd_line_tools:visa_info
+
+[options.package_data]
+* = *.dll
```
MatthieuDartiailh: Why is this line here we do not ship any pre-built dll ?

KOLANICH: I have only translated `setup.py` + some minor changes related to packaging. I also think that wheels should contain no dlls. But I estimated that the chances are higher to get this PR merged if it is not changing anything that might be controversal. So  I haven't removed the dlls. Especially since I cannot test it now since I'm on a linux machine.

MatthieuDartiailh: My bad I have some fake dll files to test reading the headers. So it should stay. I will merge soon.

Issue #506, opened 2020-05-01
pyvisa no longer able to load Keysight VISA in Python 3.8 on Windows


hognala:
Python 3.8 changes .dll load behavior. See https://github.com/Toblerity/Fiona/issues/851

This used to work:
```
import pyvisa
rm = pyvisa.ResourceManager('ktvisa32')
```

... but now it doesn't. The problem is that Keysight VISA stores a bunch of its .dll's in a separate directory (not c:\windows\system32, but C:\Program Files (x86)\Keysight\IO Libraries Suite\bin) that is on the PATH. With the change in behavior, the PATH is no longer searched and so the ktvisa32.dll .dll fails to load because its dependant .dll's can't be found.

As a workaround, users can use the following syntax:

**32-bit Python:**
```
import os
os.add_dll_directory('C:\\Program Files (x86)\\Keysight\\IO Libraries Suite\\bin')
import pyvisa
rm = pyvisa.ResourceManager('ktvisa32')
```

**64-bit Python:**
```
import os
os.add_dll_directory('C:\\Program Files\\Keysight\\IO Libraries Suite\\bin')
import pyvisa
rm = pyvisa.ResourceManager('ktvisa32')
```
 
Feel free to reach out to me if you need someone to test a fix, or something (alan_copeland@keysight.com)

MatthieuDartiailh:
Thanks for the report. I must say I missed this change in Python 3.8.

We can attack this in multiple ways. The first is simply to document this new behavior since the fix is quite straightforward. The second, which goes with this the first, is to add the possibility to specify additional dll path in the pyvisa configuration file. Those two are low maintenance and apply to more than Keysight.
A third thing we could is to systematically add the Keysight folder on the path if it exists and we are on Windows.

Out of curiosity, what is the usecase that requires to specifically load the Keysight VISA in this way rather than the generic VISA.

If you can provide a patch that would be great. Otherwise it may need to wait a couple of weeks since I am finishing a HUGE update (code style, dosctring, format, linting, typing), those things should not have ended up being bunched together but it is too far along to stop. If you are available to test the patch once it is ready I would appreciate it too.

hognala:
Option 2 seems to make the most sense (extra search paths in configuration file).

Do you have standard configuration files that you use for the different vendor's or just the info that is in the documentation? 

[...]

PR #509, opened 2020-05-06
Add the ability to specify 'dependency paths' which are used during .…

hognala:
…dll dependency resolution.

I tested that this works with `visa library` defined, and when it is not. Also, if this section is not present the .pyvisarc file still loads.

The attached file will allow the latest version of Keysight VISA to work with Python 3.8, even without specifying the .dll (it allows it to find its dependency .dll's).

[.pyvisarc.zip](https://github.com/pyvisa/pyvisa/files/4584801/default.pyvisarc.zip)

### DIFF
--- a/docs/source/introduction/configuring.rst
+++ b/docs/source/introduction/configuring.rst
@@ -66,11 +66,11 @@ contain the following::
    [Paths]
 
    VISA library: /usr/lib/libvisa.so.7
-   dependency paths=/my/otherpath/;/my/otherpath2
+   dll_extra_paths=/my/otherpath/;/my/otherpath2
 
 Please note that `[Paths]` is treated case-sensitively.
 
-`dependency paths` can be used to specify folders to search for library dependencies
+`dll_extra_paths` can be used to specify folders to search for library dependencies
 
 You can define a site-wide configuration file at
 :file:`/usr/share/pyvisa/.pyvisarc` (It may also be
--- a/docs/source/introduction/configuring.rst
+++ b/docs/source/introduction/configuring.rst
@@ -66,11 +66,15 @@ contain the following::
    [Paths]
 
    VISA library: /usr/lib/libvisa.so.7
-   dll_extra_paths=/my/otherpath/;/my/otherpath2
 
 Please note that `[Paths]` is treated case-sensitively.
 
-`dll_extra_paths` can be used to specify folders to search for library dependencies
+To specify extra .dll search paths on Windows, use a `.pyvisarc` file like the following::
+
+   [Paths]
+
+   dll_extra_paths: C:\Program Files\Keysight\IO Libraries Suite\bin;C:\Program Files (x86)\Keysight\IO Libraries Suite\bin
+
 
 You can define a site-wide configuration file at
 :file:`/usr/share/pyvisa/.pyvisarc` (It may also be
--- a/docs/source/introduction/configuring.rst
+++ b/docs/source/introduction/configuring.rst
@@ -66,8 +66,11 @@ contain the following::
    [Paths]
 
    VISA library: /usr/lib/libvisa.so.7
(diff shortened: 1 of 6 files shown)


hognala:
Fixes issue 506 - https://github.com/pyvisa/pyvisa/issues/506

docs/source/introduction/configuring.rst
```diff
@@ -66,9 +66,12 @@ contain the following::
    [Paths]
 
    VISA library: /usr/lib/libvisa.so.7
+   dependency paths=/my/otherpath/;/my/otherpath2
```
MatthieuDartiailh, line 69: Since this is windows specific I would use a more descriptive name such as (I think using dll in the name will clarify the impact):
- dll paths
- dll extra paths
or similar

pyvisa/util.py
```diff
@@ -82,6 +83,12 @@ def read_user_library_path():
 
     logger.debug('User defined library files: %s' % files)
     try:
+        try:
+            dependencyPaths = config_parser.get("Paths", "dependency paths")
+            for path in dependencyPaths.split(";"):
+                os.add_dll_directory(path)
```
MatthieuDartiailh, line 89: This is available only on windows (os.add_dll_directory docs in https://docs.python.org/3/library/os.html )so I think this whole section should be behind a `sys.platform` check. I can easily see people copy pasting a file from a windows to a linux machine and I would prefer to avoid crahsing because the function is not available.

docs/source/introduction/configuring.rst
```diff
@@ -66,9 +66,12 @@ contain the following::
    [Paths]
```
MatthieuDartiailh, line 66: We should have a note warning people about the change of behavior in Python 3.8 and how to modify their code to get things working again. We can then explain how to persist this option in the config file.

pyvisa/util.py
```diff
@@ -64,6 +64,7 @@ def read_user_library_path():
```
MatthieuDartiailh, line 64: Loading the DLL extra paths when reading the user library path could lead to weird situation since it will be called after looking for standard names (visa, visa32, etc). I think it is unlikely that somebody on windows would manage to get in a situation where VISA is not where we expect it but I think it would be cleaner to have a new utility function called in IVIVISALibrary.get_library_paths before scanning for any dll just in case.

docs/source/introduction/configuring.rst
hognala: Makes sense, I'll use dll_extra_paths

pyvisa/util.py
hognala: Easy enough, makes sense

docs/source/introduction/configuring.rst
hognala: Do you have a style of 'comment' to follow? This one feels more squishy...

pyvisa/util.py
hognala: Good point, I'll look at that.

docs/source/introduction/configuring.rst
hognala: I took a stab at updating the FAQ and the Configuring the backend section

hognala:
@MatthieuDartiailh - all comments resolved, and unittests passing now

MatthieuDartiailh:
Thanks for the ping. I will review in details and merge this weekend.

MatthieuDartiailh:
I took the liberty to update the changelog and modify the tests a bit to avoid them leaking monkeypatched value.

Thanks for your contribution

@bors r+

MarekPetr:
I still have the

> Fixes issue 506 - #506

Unfortunately I still have this issue. https://github.com/pyvisa/pyvisa/issues/506#issuecomment-762248678

PR #527, opened 2020-07-23
docs: mention that ResourceManager.list_resources() may not return open devices

itayperl:
(Related to pyvisa/pyvisa-py#244)


MatthieuDartiailh:
The error is unrelated to your changes. I think we could still add a note in pyvisa-py (under the USB section of installation for lack of a better place maybe).

PR #598, opened 2021-05-11
ctwrapper: fix debug info collection

MatthieuDartiailh:
Correct the information for ctypes debug info (previously we were getting the manufacturer name 3 times !)


MatthieuDartiailh:
The bot is currently down but the tests are passing locally. Given that the bot situation may last a bit, merging now.

PR #651, opened 2022-01-24
change the copyright date to 2022

LongnoseRob:
A quick change of all copyright dates to -2022.

Came across this on another project and also did the small change for pyvisa

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2005-2019 PyVISA Authors and contributors. See AUTHORS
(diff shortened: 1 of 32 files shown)


MatthieuDartiailh:
I am never clear on when a copyright should be updated, usually I only touch it when touching the file. Do you have a reference about when it makes sense to update it ?

LongnoseRob:
From my point of view the repository should have the copyright date including the current year.
But it might also be valid only to push the copyright date if a release is prepared or even if the files are touched.

It seems there are also some ways of automating such work on GH.

Issue #640, opened 2021-12-01
[COM] Communication issue with HP3577A using GPIB and VISA


sixtemesseven:
I use a HP3577 connected via a genuine NI USB-GPIB-HS adapter to my host PC. It used to work last week, then I suddenly couldn't run my scripts any more. I noticed that even a minimal code:

```
import pyvisa as visa

rm = visa.ResourceManager()
print(rm)
hp = rm.open_resource('GPIB0::11::INSTR')
print(hp)
hp.write('AB4')
```

Caused the issue:

```
Resource Manager of Visa Library at C:\WINDOWS\system32\visa32.dll
GPIBInstrument at GPIB0::11::0::INSTR
Traceback (most recent call last):

  File "D:\Github_Sixtimesseven\VisaInstruments\Simplevisa\SimpleVisa\untitled0.py", line 14, in <module>
    hp.write('AB4')

  File "C:\Users\justrandom\anaconda3\envs\ba21\lib\site-packages\pyvisa\resources\messagebased.py", line 197, in write
    count = self.write_raw(message.encode(enco))

  File "C:\Users\justrandom\anaconda3\envs\ba21\lib\site-packages\pyvisa\resources\messagebased.py", line 157, in write_raw
    return self.visalib.write(self.session, message)[0]

  File "C:\Users\justrandom\anaconda3\envs\ba21\lib\site-packages\pyvisa\ctwrapper\functions.py", line 2796, in write
    ret = library.viWrite(session, data, len(data), byref(return_count))

  File "C:\Users\justrandom\anaconda3\envs\ba21\lib\site-packages\pyvisa\ctwrapper\highlevel.py", line 222, in _return_handler
    return self.handle_return_value(session, ret_value)  # type: ignore

  File "C:\Users\justrandom\anaconda3\envs\ba21\lib\site-packages\pyvisa\highlevel.py", line 251, in handle_return_value
    raise errors.VisaIOError(rv)

VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
```

Causes the instrument to go into remote and lock up (power cycle required), it does not set average mode 4 (AB4 command) which is tested and should work and has worked last week.
I tried another instrument and it was not working as well (Keithley 238). 

I tried another PC and uninstalling visa and the 488-2 drivers. I reinstalled the oldest (17.6) version and updated pyvisa (it was already up to date). Right now I can talk to the other instruments but the HP3577 locks up as before.

Instrument details
------------------
[...]

PR #653, opened 2022-02-01
GPIB improve secondary address handling

MatthieuDartiailh:
If a resource does not use a GPIB device secondary address, it should be fully omitted from the resource name rather than set to 0. Setting the secondary address to 0 can cause issues with old instruments (see #640) and also cause issues when using low-level GPIB protocols (can't find the issue ATM).

@sixtemesseven could you review and tests as you offered to do in #640


MatthieuDartiailh:
I tested this locally and it works as expected, so merging.

PR #661, opened 2022-04-03
Adding support for pyvisa_proxy #660

### LINKED ISSUE #660: [REQUEST] Support for integrating pyvisa-proxy
Hi Pyvisa Devs,

I am currently [developing an extension](https://github.com/casabre/pyvisa-proxy) for pyvisa which should somehow reflect the behavior of the NI remote VISA functionality but without the constraint of running NI-VISA. Due to the fact that there are many VISA flavors out there and NI-VISA is a bloated installer, I want to develop an independent and generic solution based on zmq which are connecting legacy protocols as GPIB or some USB connection. In the end, with mostly and ethernet-based instruments and pyvisa-proxy, distributed, dockerized tests will be feasible like for instance with my corresponding [Docker Image](https://hub.docker.com/r/casabre/rsvisa-python).

The architecture is not thought to then end but I just want to see if the PoC is working. However, after starting, I realized that implementing the client side is pretty hard because I want to forward the attribute calls to the server. Without relying on a specific class implementation, it was almost impossible to get the right attributes for forwarding.

Thus, I am asking if you could support me in order to integrate the logic into pyvisa with either a more monkey-patching approach of `__get_attr__` or an embedded implementation. 

Regarding the resource string, I have something similar in mind like the NI-VISA string &rarr; `proxy://hostname:port/remote resource`. After parsing, the pyvisa-proxy extension has to be loaded.

If that sounds interesting to you and I believe you have questions, please contact me.

Thanks & best regards
Carsten

casabre:
- [x] Closes #660

### DIFF
--- a/docs/source/advanced/backends.rst
+++ b/docs/source/advanced/backends.rst
@@ -91,6 +91,11 @@ As a **very minimum** set you need:
 
 (you can get the signature below or here :ref:`api_visalibrarybase`)
 
+.. note::
+A |ResourceManager| attribute check for 
+:class:`pyvisa.highlevel.VisaLibraryBase.open_resource` is implemented 
+in order to support custom opening handling by the new backend implementation.
+
 But of course you cannot do anything interesting with just this. In general you
 will also need:
 
@@ -194,7 +199,3 @@ Complete list of level 2 functions to implement::
     def write(self, session, data):
     def write_asynchronously(self, session, data):
     def write_from_file(self, session, filename, count):
-
-Additionally, a |ResourceManager| attribute check for 
-:class:`pyvisa.highlevel.VisaLibraryBase.open_resource` is implemented 
-in order to support custom opening handling by the new backend implementation.
--- a/docs/source/advanced/backends.rst
+++ b/docs/source/advanced/backends.rst
@@ -194,3 +194,7 @@ Complete list of level 2 functions to implement::
     def write(self, session, data):
     def write_asynchronously(self, session, data):
     def write_from_file(self, session, filename, count):
+
+Additionally, a |ResourceManager| attribute check for 
+:class:`pyvisa.highlevel.VisaLibraryBase.open_resource` is implemented 
+in order to support custom opening handling by the new backend implementation.
--- a/docs/source/advanced/backends.rst
+++ b/docs/source/advanced/backends.rst
@@ -91,10 +91,11 @@ As a **very minimum** set you need:
 
 (you can get the signature below or here :ref:`api_visalibrarybase`)
 
-.. note::
-A |ResourceManager| attribute check for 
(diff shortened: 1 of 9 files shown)


MatthieuDartiailh:
Sorry for the long delay. This looks reasonable. Could you fix the formatting issue and add an entry to the changelog and  to the documentation ?

casabre:
@MatthieuDartiailh thanks for update and no worries, I was also quite busy the last weeks.

Regarding the CI issues, I fixed my parts but they were raised at spots I didn't change at all. They came from main. I can fix them but I don't want to be too intrusive in your code base 😉.

Furthermore, I want to add also an additional unit test which covers the new behavior.

If it is merged and a new version published, I will refer to that version as a minimum version.

MatthieuDartiailh:
The CIs failure is likely a new black thing feel free to fix it in a dedicated commit.

casabre:
@MatthieuDartiailh I fixed the black issues and maybe they were already fixed at upstream. Furthermore, I have added a test for the open_resource attribute check and updated the CHANGES file. 

Regarding the documentation, I don't know where to put some information. Thus, I am seeing no need to extend it at the moment.

MatthieuDartiailh (approved):

MatthieuDartiailh:
Formatting still looks unhappy...

Also I still think some documentation in https://github.com/pyvisa/pyvisa/blob/main/docs/source/advanced/backends.rst would be good.

casabre:
Hm, I am really not getting it. This is my local result.
```shell
$ black pyvisa --check
All done! ✨ 🍰 ✨
50 files would be left unchanged.
```
I am having the latest Ubuntu 22.04 with Python 3.10.4

MatthieuDartiailh:
The error comes from isort not black

casabre:
Sorry for the confusion. I have picked the wrong CI results. Pushed now another `isort` run and extended the backend documentation.

docs/source/advanced/backends.rst
```diff
@@ -194,3 +194,7 @@ Complete list of level 2 functions to implement::
     def write(self, session, data):
     def write_asynchronously(self, session, data):
     def write_from_file(self, session, filename, count):
+
+Additionally, a |ResourceManager| attribute check for 
+:class:`pyvisa.highlevel.VisaLibraryBase.open_resource` is implemented 
+in order to support custom opening handling by the new backend implementation.
```
MatthieuDartiailh, line 200: I think this would be easier to discover as a `.. note::` directive on line 93 since people checking what they need to implement are likely to be interested in this.

casabre: Sure if it increase visibility! Just moved it to line 93.

MatthieuDartiailh:
All good. 

I have  a bunch of things I need to finish before the release and since now you will be waiting on it I will try to prioritize them. In case I need it would you be able/willing to test some patches before I merge them.

PR #687, opened 2022-08-17
Remove duplicate call to shell.main

sean-moore3:

Removes a double call to `shell.main` in `cmd_line_tools.py`.

Before the change, calling `exit` from `pyvisa-shell` results in a re-entry of the command loop.  `exit` has to be called twice.

image

After this change:

image


MatthieuDartiailh:
Thanks

PR #688, opened 2022-08-18
fixed a couple bugs around is_64bits

bobmcnamara:
If these aren't bugs, they are at least deserving of a comment.


MatthieuDartiailh:
I believe you are right and those are bugs.

PR #697, opened 2022-10-05
Bug fixes related to get_visa_attribute()

bobmcnamara:
I found some bugs while comparing how the Rohde & Schwarz VISA library responds to get_visa_attribute calls, versus the pyvisa-py VISA library.

I didn't understand why some of the constants.ResourceAttribute members weren't defined in the attributes.AttributesByID dict, but generating VI_ERROR_NSUP_ATTR in those cases produces the desired behavior.

### DIFF
--- a/pyvisa/attributes.py
+++ b/pyvisa/attributes.py
@@ -390,23 +390,20 @@ class CharAttribute(Attribute):
 # Attributes are in the same order as in the constants.ResourceAttribute enum
 
 
-# class AttrVI_ATTR_RM_SESSION(Attribute):
-#     """Specifies the session of the Resource Manager used to open this session."""
-#
-#     resources = AllSessionTypes
-#
-#     py_name = "resource_manager_session"
-#
-#     visa_name = "VI_ATTR_RM_SESSION"
-#
-#     visa_type = "ViSession"
-#
-#     default = NotAvailable
-#
-#     read, write, local = True, False, False
-#
-# VI_ATTR_RM_SESSION is not implemented as resource property,
-# use .resource_manager.session instead
+class AttrVI_ATTR_RM_SESSION(Attribute):
+    """Specifies the session of the Resource Manager used to open this session."""
+
+    resources = AllSessionTypes
+
+    py_name = ""
+
+    visa_name = "VI_ATTR_RM_SESSION"
+
+    visa_type = "ViSession"
+
+    default = NotAvailable
+
+    read, write, local = True, False, False
 
 
 class AttrVI_ATTR_INTF_TYPE(EnumAttribute):
--- a/pyvisa/attributes.py
+++ b/pyvisa/attributes.py
@@ -390,22 +390,21 @@ class CharAttribute(Attribute):
 # Attributes are in the same order as in the constants.ResourceAttribute enum
 
 
-class AttrVI_ATTR_RM_SESSION(Attribute):
-    """Specifies the session of the Resource Manager used to open this session."""
-
-    resources = AllSessionTypes
-
-    py_name = "resource_manager_session"
-
-    visa_name = "VI_ATTR_RM_SESSION"
-
-    visa_type = "ViSession"
-
-    default = NotAvailable
-
-    read, write, local = True, False, False
-
-
+# class AttrVI_ATTR_RM_SESSION(Attribute):
+#     """Specifies the session of the Resource Manager used to open this session."""
+#
+#     resources = AllSessionTypes
+#
+#     py_name = "resource_manager_session"
+#
+#     visa_name = "VI_ATTR_RM_SESSION"
+#
+#     visa_type = "ViSession"
+#
+#     default = NotAvailable
+#
+#     read, write, local = True, False, False
+#
 # VI_ATTR_RM_SESSION is not implemented as resource property,
 # use .resource_manager.session instead
 
--- a/pyvisa/attributes.py
+++ b/pyvisa/attributes.py
@@ -1292,7 +1292,7 @@ class AttrVI_ATTR_TCPIP_HISLIP_MAX_MESSAGE_KB(RangeAttribute):
 
     visa_name = "VI_ATTR_TCPIP_HISLIP_MAX_MESSAGE_KB"
 
-    visa_type = "ViUint32"
+    visa_type = "ViUInt32"
 
     default = 1024
 
@@ -2611,7 +2611,7 @@ class AttrVI_ATTR_WIN_SIZE(RangeAttribute):
 
     visa_name = "VI_ATTR_WIN_SIZE"
 
-    visa_type = "ViBusSize64" if constants.is_64bits else "ViBusSize"
+    visa_type = "ViUInt64" if constants.is_64bits else "ViUInt32"
 
     default = NotAvailable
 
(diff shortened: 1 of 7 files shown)


pyvisa/attributes.py
```diff
@@ -2611,7 +2611,7 @@ class AttrVI_ATTR_WIN_SIZE(RangeAttribute):
 
     visa_name = "VI_ATTR_WIN_SIZE"
 
-    visa_type = "ViBusSize64" if constants.is_64bits else "ViBusSize"
```
MatthieuDartiailh, line 2614: The ViBusSize64 type exists according to the VISA specs so I would prefer to add it in types.py rather than using a type differing from the specs even if they all just alias one another.

pyvisa/ctwrapper/functions.py
```diff
@@ -832,6 +832,19 @@ def get_attribute(library, session, attribute):
...
     """
+    if attribute not in attributes.AttributesByID:
+        #
+        # the following ResourceAttribute members will fail:
+        #
+        #     fdc_generate_signal_enabled
+        #     interface_parent_number
+        #     pxi_memory_base_bar5
+        #     pxi_memory_size_bar5
+        #     pxi_memory_type_bar5
```
MatthieuDartiailh, line 843: Those can be recovered by expanding the range to found on line 3720 of attributes to be `range(0, 6)`

pyvisa/ctwrapper/functions.py
```diff
@@ -832,6 +832,19 @@ def get_attribute(library, session, attribute):
         Return value of the library call.
 
     """
+    if attribute not in attributes.AttributesByID:
```
MatthieuDartiailh, line 835: This should not be needed and I would prefer to add the missing attribute definitions. I would also be curious to know how you found those and if this is something that could be integrated in the testsuite ?

MatthieuDartiailh (changes requested):

pyvisa/ctwrapper/functions.py
bobmcnamara: `from pyvisa.constants import ResourceAttribute
from pyvisa import VisaIOError
import pyvisa

rm = pyvisa.ResourceManager("@py")
scope = rm.open_resource("TCPIP::localhost::hislip0")

for attr in dir(ResourceAttribute):
    if attr.startswith("_"):
        continue
    try:
        value = scope.get_visa_attribute(getattr(ResourceAttribute, attr))
    except VisaIOError as e:
        value = str(e)
        if value.startswith("VI_ERROR_NSUP_ATTR"):
            continue
    except AttributeError as e:
        value = repr(e)
    except KeyError as e:
        value = repr(e)

    print(f"{attr}: {value!r}")
`
Maybe there could be a test case to verify that for all attributes in ResourceAttribute, get_visa_attribute() either returns a value or raises VisaIOError with an error message that starts with "VI_ERROR_NSUP_ATTR".

MatthieuDartiailh: We could add such a test but that requires a fake instrument and the dedicated worker is not in shape to run it and I do not have the bandwidth to fix it ATM.

pyvisa/attributes.py
bobmcnamara: Fixed.

pyvisa/ctwrapper/functions.py
bobmcnamara: Fixed.

pyvisa/ctwrapper/functions.py
bobmcnamara: Fixed.

pyvisa/resources/resource.py
```diff
@@ -208,6 +208,9 @@ def resource_info(self) -> highlevel.ResourceInfo:
     #: VISA attributes require the resource to be opened in order to get accessed.
     #: Please have a look at the attributes definition for more details
 
+    #: The session of the Resource Manager used to open this session.
+    resource_manager_session: Attribute[int] = attributes.AttrVI_ATTR_RM_SESSION()
+
```
MatthieuDartiailh, line 213: Since the same information is accessible through `resource_manager.session` I think it is not worth exposing it here. Having it define allow to have a uniform behavior for `get_attribute` which is important but exposing it would just be confusing IMO.

pyvisa/attributes.py
```diff
@@ -389,6 +389,23 @@ def pre_set(self, value: Union[str, bytes]) -> int:
 # --- Session attributes ---------------------------------------------------------------
 # Attributes are in the same order as in the constants.ResourceAttribute enum
 
+
+class AttrVI_ATTR_RM_SESSION(Attribute):
+    """Specifies the session of the Resource Manager used to open this session."""
+
+    resources = AllSessionTypes
+
+    py_name = "resource_manager_session"
```
MatthieuDartiailh, line 398: If we do not expose it there is no value in setting the py_name.

MatthieuDartiailh (changes requested):
Thanks a lot for your work and the typo fixes. I think this is nearly good to go. I would simply revert the direct exposition of resource_manager_session which has an existing equivalent.

pyvisa/attributes.py
```diff
@@ -389,6 +389,22 @@ def pre_set(self, value: Union[str, bytes]) -> int:
...
+#     py_name = "resource_manager_session"
+#
+#     visa_name = "VI_ATTR_RM_SESSION"
+#
+#     visa_type = "ViSession"
+#
+#     default = NotAvailable
+#
+#     read, write, local = True, False, False
+#
```
MatthieuDartiailh, line 407: Am I wrong in thinking that if you fully remove it you will run into your first issue will re-appear ? I think that we want to allow people that want to use get_visa_attribute to use it but just not provide multiple high-level ways to do the same thing.

So I would un-comment this (and clear the py_name).

pyvisa/constants.py
```diff
@@ -1900,7 +1900,10 @@ class ResourceAttribute(enum.IntEnum):
     """The possible attributes of VISA resources."""
 
     # All sessions
-    resource_manager_session = VI_ATTR_RM_SESSION
+
+    ## VI_ATTR_RM_SESSION is not implemented as resource property,
+    ## use .resource_manager.session instead
+    # resource_manager_session = VI_ATTR_RM_SESSION
```
MatthieuDartiailh, line 1906: This can be reverted.

pyvisa/resources/resource.py
```diff
@@ -208,6 +208,11 @@ def resource_info(self) -> highlevel.ResourceInfo:
     #: VISA attributes require the resource to be opened in order to get accessed.
     #: Please have a look at the attributes definition for more details
 
+    ## VI_ATTR_RM_SESSION is not implemented as resource property,
+    ## use .resource_manager.session instead
+    #: The session of the Resource Manager used to open this session.
+    # resource_manager_session: Attribute[int] = attributes.AttrVI_ATTR_RM_SESSION()
+
```
MatthieuDartiailh, line 215:
```suggestion
    # VI_ATTR_RM_SESSION is not implemented as resource property,
    # use .resource_manager.session instead
    # resource_manager_session: Attribute[int] = attributes.AttrVI_ATTR_RM_SESSION()

```

MatthieuDartiailh (commented):
It seems my last comments were confusing:
- I am in favor of supportinh get_visa_attribute for all attributes
- I am against cluttering the API with direct descriptor accesses for things for which we provide already a simple API (because we can bypass VISA)

MatthieuDartiailh:
Can you add a changelog entry and I will merge ?

MatthieuDartiailh:
Thanks again for your contribution

PR #727, opened 2023-02-24
read_termination_context uses read_termination logic

### LINKED ISSUE #723: read_termination_context does not support None
`MessageBasedResource.read_termination_context` requires a string and does not support an empty string or None. Due to the slicing `[-1]`:
https://github.com/pyvisa/pyvisa/blob/c76286ff4bfca1f858d1316d2e6a0916f8e6bdac/pyvisa/resources/messagebased.py#L775-L782

As a solution, the contextmanager could use the `read_termination` property (which checks, whether the read_termination exists) or copy the code from the read_termination into the contextmanager.

If you want, I might contribute the according pull request.

**Output of `pyvisa-info`**
Machine Details:
   Platform ID:    Windows-10-10.0.19045-SP0
   Processor:      Intel64 Family 6 Model 158 Stepping 11, GenuineIntel

Python:
   Implementation: CPython
   Executable:     C:\Users\moneke.NLOQO\AppData\Local\Programs\Python\Python310\python.exe
   Version:        3.10.5
   Compiler:       MSC v.1929 64 bit (AMD64)
   Bits:           64bit
   Build:          Jun  6 2022 16:14:13 (#tags/v3.10.5:f377153)
   Unicode:        UCS4

PyVISA Version: 1.11.3

Backends:
   ivi:
      Version: 1.11.3 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.5.2
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR:
         PyUSB does not seem to be properly installed.
         Please refer to PyUSB documentation and
         install a suitable backend like
         libusb 0.1, libusb 1.0, libusbx,
         libusb-win32 or OpenUSB.
         No backend available
      USB RAW:
         PyUSB does not seem to be properly installed.
         Please refer to PyUSB documentation and
         install a suitable backend like
         libusb 0.1, libusb 1.0, libusbx,
         libusb-win32 or OpenUSB.
         No backend available
      TCPIP INSTR: Available
      TCPIP SOCKET: Available
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of funcionality.
         No module named 'gpib'
   sim:
      Version: 0.4.0
      Spec version: 1.1

bmoneke:
Problem was, that the read_termination_context did not accept None for no read_termination.

Changes:
- Logic of `read_termination.setter` moved to a new method (without setting private attribute `_read_termination`).
- `read_termination_context` uses that new method to set the termination before and after yield.

Open question:
Right now, the private attribute `_read_termination` (returned by public property `read_termination`) remains the original one.
If you prefer, we could change that, such that the method looks this:
~~~python
def read_termination_context(self, value):
    term = self.read_termination
    yield
    self.read_termination = term
~~~

- [x] Closes #723 (insert issue number if relevant)
- [ ] The change is fully covered by automated unit tests (there are no test_files for `messagebased` yet)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA Changelog
 1.14.0 (unreleased)
 -------------------
 - fix ctypes truncated pointers on 64-bit for ViBusAddress, ViBusSize, ViAttrState PR # 725
+- fix read_termination_context does not allow None PR # 727
 
 1.13.0 (22-12-2022)
 -------------------
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -773,12 +773,10 @@ class MessageBasedResource(Resource):
 
     @contextlib.contextmanager
     def read_termination_context(self, new_termination: str) -> Iterator:
-        term = self.get_visa_attribute(constants.ResourceAttribute.termchar)
-        self.set_visa_attribute(
-            constants.ResourceAttribute.termchar, ord(new_termination[-1])
-        )
+        term = self.read_termination
+        self.read_termination = new_termination
         yield
-        self.set_visa_attribute(constants.ResourceAttribute.termchar, term)
+        self.read_termination = term
 
     def flush(self, mask: constants.BufferOperation) -> None:
         """Manually clears the specified buffers.


MatthieuDartiailh:
I dislike the idea of read_termination not providing access to the value currently used. So I would be in favor of updating it in the context manager. Also please try to fix the CI failures.

bmoneke:
Ok, I reverted the change to the property (not needed anymore) and based the context_manager on the `read_termination` property, such that the context manager does, what you expect: setting it beforehand to some temporary value and returning it to the previous value afterwards.

bmoneke:
Apparently it wants to reformat a bunch of other files.
This should not enter this PR though, I guess.

MatthieuDartiailh:
I updated main, could you rebase your PR ?

bmoneke:
Done.
Thanks.

MatthieuDartiailh:
Thanks

PR #733, opened 2023-03-26
Use architecture instead of `bitness` in `LibraryPath` and friends

### LINKED ISSUE #655: Macbook , They are all 64bits, but tips 'no matching architecture in universal wrapper'
Machine Details:
   Platform ID:    macOS-11.2.3-arm64-arm-64bit
   Processor:      arm

Python:
   Implementation: CPython
   Executable:     /opt/homebrew/opt/python@3.9/bin/python3.9
   Version:        3.9.9
   Compiler:       Clang 13.0.0 (clang-1300.0.29.3)
   Bits:           64bit
   Build:          Nov 21 2021 03:16:13 (#main)
   Unicode:        UCS4

PyVISA Version: 1.11.3

Backends:
   ivi:
      Version: 1.11.3 (bundled with PyVISA)
      #1: /Library/Frameworks/visa.framework/visa:
         found by: auto
         bitness: 64
         Could not get more info: Interpreter and library have different bitness.

import pyvisa
rm = pyvisa.ResourceManager('/Library/Frameworks/VISA.framework/VISA')
rm.list_resources()

OSError: dlopen(/Library/Frameworks/VISA.framework/VISA, 6): no suitable image found.  Did find:
	/Library/Frameworks/VISA.framework/VISA: no matching architecture in universal wrapper
	/Library/Frameworks/VISA.framework/VISA: no matching architecture in universal wrapper
	/Library/Frameworks/VISA.framework/Versions/A/VISA: no matching architecture in universal wrapper

arr-ee:
- Replaced `bitness` of libraries and python interpreter
- Removed unused `LibraryPath.is32bit`, `LibraryPath.is_64bit`, and `LibraryPath.bitness` methods
- Introduced AARCH64 as tested-for architecture
- Slightly updated detection logic
- Removed IA64 and linux universal binary test scenarios since they are unrealistic
- Removed `linux2/3` checking since it should not exist past Python 3.3

I am not entirely happy with current enum-based approach, but I still think enums are more appropriate in this case than magic dicts. Maybe we can re-use (renamed) PEMachineType for arch info.

I was also considering having actual fake binaries for all platforms/architectures we test for stored in test artifacts since current approach with mocking is quite brittle.

Please let me know if you had a different way of tacking this in mind.

- [x] Closes #655 (insert issue number if relevant)

### DIFF
--- a/pyvisa/testsuite/test_util.py
+++ b/pyvisa/testsuite/test_util.py
@@ -14,6 +14,7 @@ import unittest
 from configparser import ConfigParser
 from functools import partial
 from io import StringIO
+from pathlib import Path
 from types import ModuleType
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


arr-ee:
I'm not sure why mypy suddenly decided to act on an issue that appears to have been there for some 14 months or so. Am I missing something, or shall I open an issue to address it?

MatthieuDartiailh (commented):
This looks nice. Could you fix  the tests and explain to me why you switched from tmpdir to tmp_path (for my own curiosity) ?

MatthieuDartiailh:
Mypy has become stricter recently. Feel free to address the issue in this PR.

arr-ee:
> explain to me why you switched from tmpdir to tmp_path (for my own curiosity)

Apologies for that change, it's more of a bugbear of mine. `Path` provides a surprisingly ergonomic (not to mention type-checked and reasonably cross-platform) API for working with file system, and I've been habitually updating the codebase as I was going with no regard to introducing inconsistencies. I'm OK with either reverting that part of the changeset or filing a follow-up to update the rest of the lib — up to you.

While we're talking about builds suddenly failing due to stealth mypy upgrade: any opinions on `pipenv`/another way to have lock-files for dependencies?

Thanks!

MatthieuDartiailh:
I am fine with the change. I must have missed the introduction of that fixture (I remember being bitten often enough by tmpdir that I do not mind the change).

Honestly, I prefer mypy flagging incorrect/dubious things. I can take care of it if you want, but having to update pipenv myself would likely result in quite a lag and I prefer to avoid that.

pyvisa/testsuite/test_util.py
```diff
@@ -509,80 +510,84 @@ def get_debug_info(cls):
...
+            ["x86_32", "x86_64", "arm_64"],
+            [
+                util.PEMachineType.I386,
+                util.PEMachineType.AMD64,
+                util.PEMachineType.AARCH64,
+            ],
+        ):
             arch = util.get_shared_library_arch(
-                os.path.join(tmpdir, "fakelib_good%s.dll" % f)
+                Path(tmp_path, "fakelib_good_%s.dll" % f)
```
MatthieuDartiailh, line 556: Since tmp_path is a Path, simply use the / based syntax.

arr-ee: oh, good call

pyvisa/util.py
```diff
@@ -171,7 +223,7 @@ class LibraryPath(str):
     found_by: str
 
     #: Architectural information (32, ) or (64, ) or (32, 64)
```
MatthieuDartiailh, line 225: Please update the documentation comment

pyvisa/util.py
```diff
@@ -930,49 +968,15 @@ def get_debug_info(to_screen: Literal[False]) -> str:
...
-    0x0CEF: "CEF",
-    0x0EBC: "EBC",
-    0x8664: "AMD64",
-    0x9041: "M32R",
-    0xC0EE: "CEE",
-}
-
-
-def get_shared_library_arch(filename: str) -> str:
+def get_shared_library_arch(filename: Path) -> PEMachineType:
```
MatthieuDartiailh, line 979: The type annotation is overly restrictive here
```suggestion
def get_shared_library_arch(filename: str | Path) -> PEMachineType:
```
or get the annotation from io

pyvisa/util.py
```diff
@@ -991,35 +995,64 @@ def get_shared_library_arch(filename: str) -> str:
...
 
-        return machine_types.get(machine, "UNKNOWN")
+        try:
+            return PEMachineType(machine)
+        except ValueError:
+            return PEMachineType.UNKNOWN
 
 
-def get_arch(filename: str) -> Tuple[int, ...]:
+def get_arch(filename: Path) -> List[ArchitectureType]:
```
MatthieuDartiailh, line 1004: Same as above

pyvisa/util.py
```diff
@@ -183,37 +235,16 @@ def __new__(
...
-            return "n/a"
-        return 64 in self.arch
-
-    @property
-    def bitness(self) -> str:
-        """Bitness of the library."""
-        if not self.arch:
-            return "n/a"
-        return ", ".join(str(a) for a in self.arch)
-
```
MatthieuDartiailh, line 216: This is technically backward incompatible... But it is far from the core so we can ride with it.

Can you please add an entry to the changelog mentioning that those were removed.

MatthieuDartiailh (changes requested):

MatthieuDartiailh (approved):

PR #734, opened 2023-03-26
Fix dataclass hierarchy uncovered by mypy

arr-ee:
Split out of https://github.com/pyvisa/pyvisa/pull/733 since fix ended up being a bit involved.

`mypy` was complaining about bound typevar https://github.com/pyvisa/pyvisa/blob/df50969b3527ccfa28ae7e67b2be7468865bd3e2/pyvisa/rname.py#L99 being too loose, which was true. However, we couldn't just make `ResourceName` a dataclass since it'd bring a bunch of unneeded fields into children, which was breaking all of the logics.

I ended up separating things that needed to not be in a dataclass into a separate private parent class, with `ResourceName` becoming a dataclass as it should've been.

There are opportunities to make it a bit more elegant, but I'd rather discuss that separately.

### DIFF
--- a/pyvisa/rname.py
+++ b/pyvisa/rname.py
@@ -5,7 +5,6 @@
 :license: MIT, see LICENSE for more details.
 
 """
-from __future__ import annotations
 
 import contextlib
 import re
@@ -53,7 +52,7 @@ class InvalidResourceName(ValueError):
     @classmethod
     def bad_syntax(
         cls, syntax: str, resource_name: str, ex: Optional[Exception] = None
-    ) -> InvalidResourceName:
+    ) -> Self:
         """Build an exception when the resource name cannot be parsed."""
         if ex:
             msg = "The syntax is '%s' (%s)." % (syntax, ex)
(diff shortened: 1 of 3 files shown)


arr-ee:
Oh, and pre-commit checks are now in sync with CI ;)

pyvisa/rname.py
```diff
@@ -5,10 +5,12 @@
 :license: MIT, see LICENSE for more details.
 
 """
+from __future__ import annotations
```
MatthieuDartiailh, line 8: Please don't include that future. It is rather unclear if it will ever become the default.

MatthieuDartiailh (commented):
Apart from my dislike of future annotations, it looks good.

pyvisa/rname.py
arr-ee: Sure thing; I'm pretty sure it's not that necessary at this point. I'll clean up a couple of unrelated changes as well.

MatthieuDartiailh:
You need to quote again uses of `Resource`

arr-ee:
Yeah turns out not running dev venv on the lowest supported version is not a good idea — fixed.

PR #740, opened 2023-04-14
resources: Update context manager type annotation to preserve derived type

### LINKED ISSUE #739: Type annotation for Resource context manager implicitly upcasts derived class to base class
The type annotation for `Resource.__enter__()` indicates that it returns the base class, `Resource`:
```
    def __enter__(self) -> "Resource":
        return self
```

If you narrow the resource type to a derived class before entering the context manager, entering the context manager casts it back to the base class.

[PEP 673 - Self Type](https://peps.python.org/pep-0673/) defines a language feature to support this self-return use case, which was implemented in Python 3.11. For older Python versions, you can use a type var.

To Reproduce
--------------
1. Run mypy on this code:
```
import pyvisa
import pyvisa.resources
import pyvisa.typing

def open_message_based_resource(resource_name: str) -> pyvisa.resources.MessageBasedResource:
    resource = pyvisa.ResourceManager().open_resource(resource_name)
    assert isinstance(resource, pyvisa.resources.MessageBasedResource)
    return resource

with open_message_based_resource("COM1") as r:
    print(r.query("*IDN?"))
```

Output of mypy:
```
example.py:11: error: "Resource" has no attribute "query"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

**Output of `pyvisa-info`**
```
Machine Details:
   Platform ID:    Windows-10-10.0.19044-SP0
   Processor:      AMD64 Family 23 Model 113 Stepping 0, AuthenticAMD

Python:
   Implementation: CPython
   Executable:     C:\Users\bkeryan\AppData\Local\Temp\pyvisa-mypy\.venv\Scripts\python.exe
   Version:        3.9.13
   Compiler:       MSC v.1929 64 bit (AMD64)
   Architecture:   AMD64
   Build:          May 17 2022 16:36:42 (#tags/v3.9.13:6de2ca5)
   Unicode:        UCS4

PyVISA Version: 1.13.1.dev16+g6efceab.d20230414

Backends:
   ivi:
      Version: 1.13.1.dev16+g6efceab.d20230414 (bundled with PyVISA)
      Binary library: Not found
```

bkeryan:
The type annotation for `Resource.__enter__()` indicates that it returns the base class, `Resource`. If you narrow the resource type to a derived class before entering the context manager, entering the context manager casts it back to the base class.

The solution is to use a type var to specify that the return type is the same as the type of `self`.

- [x] Closes #739 
- [ ] Executed ``black . && isort -c . && flake8`` with no errors

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,7 @@ PyVISA Changelog
     have been removed
   - `p.u.get_shared_library_arch` now returns `p.u.PEMachineType` instead of a string
   - `p.u.get_arch` now returns a list of `p.u.ArchitectureType`
+- update `Resource` context manager type annotation to preserve derived type PR # 740
 
 1.13.0 (22-12-2022)
 -------------------
--- a/pyvisa/resources/resource.py
+++ b/pyvisa/resources/resource.py
@@ -184,7 +184,7 @@ class Resource(object):
     def __repr__(self) -> str:
         return "<%r(%r)>" % (self.__class__.__name__, self._resource_name)
 
-    def __enter__(self) -> "Resource":
+    def __enter__(self: T) -> T:
         return self
 
     def __exit__(self, *args) -> None:


bkeryan:
> Executed black . && isort -c . && flake8 with no errors

The examples directory seems to have pre-existing errors:
```
(pyvisa-venv) PS C:\dev\pyvisa> isort -c .
ERROR: C:\Dev\pyvisa\examples\itc4.py Imports are incorrectly sorted and/or formatted.
ERROR: C:\Dev\pyvisa\examples\keithley2000.py Imports are incorrectly sorted and/or formatted.
(pyvisa-venv) PS C:\dev\pyvisa> flake8
.\examples\keithley2000.py:40:64: F821 undefined name 'VI_INTF_GPIB'
```

MatthieuDartiailh (commented):
Looks good. Could you update the release notes ?

MatthieuDartiailh:
Don't worry about the examples, and thanks for the fix.

PR #753, opened 2023-06-18
Cwstryker/progress bar support

cwstryker:
Added support for a tqdm (or similar) progress bar to the read_binary_values() 
and query_binary_values() methods in messagebased.py.  The monitoring interface
object (such as a tqdm progress bar) is passed to the method as an optional parameter. 

Updated the documentation and added unit test coverage.

Example Usage:
    inst = resource_manager.open_resource(...)
    ...
    with tqdm(desc="Downloading", unit="B", total=total_bytes) as progress_bar:
        inst.write("CURV?")
        data = inst.read_binary_values(..., montoring_interface=progress_bar)

Note: Unit tests were created to cover the progress bar functionality. However,
      the author does not have the Keysight virtual instrument to verify the test
      code.

- [ x ] Executed ``black . && isort -c . && flake8`` with no errors
- [ x ] The change is fully covered by automated unit tests
- [ x ] Documented in docs/ as appropriate
- [ x ] Added an entry to the CHANGES file


arr-ee:
Note: I'm not a maintainer so everything below is "for your consideration" at best.

Thank you for the PR! Glad to see the protocol-based implementation, looks good overall.

One thing that I'm not a fan of is naming: maybe `progress_cb`/`progress_hook`/another more general name would be better suited than `progress_bar` since what we're describing is a generic "number of bytes transferred" callback that can be used for other purposes.

cwstryker:
@arr-ee, you make a good point.  I refactored the API and renamed progress_bar to monitoring_interface.  I also updated the doc string.

docs/source/introduction/rvalues.rst
```diff
@@ -102,6 +102,20 @@ you get from your instrument (plus the header). If it is so, then you can
...
+the instrument. The progress bar object must be initialized with the total value
+set to the total number of bytes to be downloaded from the instrument. The progress
+bar object must implement the update() method. See the tqdm package documentation for
+more information. An implementation of a progress bar might look like:
+
+inst = resource_manager.open_resource(...)
+    ...
+    with tqdm(desc="Downloading", unit="B", total=total_bytes) as progress_bar:
+        inst.write("CURV?")
+        data = inst.read_binary_values(..., progress_bar=progress_bar)
```
arr-ee, line 117: missed rename

CHANGES
```diff
@@ -15,6 +15,8 @@ PyVISA Changelog
   - `p.u.get_shared_library_arch` now returns `p.u.PEMachineType` instead of a string
   - `p.u.get_arch` now returns a list of `p.u.ArchitectureType`
 - update `Resource` context manager type annotation to preserve derived type PR # 740
+- added support for an optional progress bar object to the read_binary_values() and
```
arr-ee, line 18: and also probably here?

arr-ee:
@cwstryker ty for a quick update! noted a couple of missed spots, otherwise it's up to @MatthieuDartiailh

cwstryker:
@arr-ee   Thanks again.  I updated the documentation in the latest commit.

MatthieuDartiailh:
The changes look fine to me. Personally I am not sure when I would use such a functionality but I guess that when dealing with some oscilloscope it could make sense. 

My bandwidth is limited but I will try to run the updated tests. Ping me if you see nothing in a week time.

cwstryker:
Yes, my target application involves large datasets downloaded from oscilloscopes.  I fixed a syntax error in the documentation and uploaded a new commit.  I will also take a look at increasing the code coverage of the test code I added and push up a new commit in the next couple of days.

MatthieuDartiailh:
Do not worry about the coverage. In an ideal world I would fix the KAT bot but I do not have the bandwidth to do so.

MatthieuDartiailh:
I had to make a couple of minor updates to get the KAT to pass. 

Honestly it makes me wonder if pyvisa could not handle the initialization of the monitor to avoid people having to juggle the issue of the headers. WDYT @cwstryker ?

cwstryker:
@MatthieuDartiailh, I agree that calculating the size of header is a bit of a hassle.  However, I don't think that read_binary_values() is right place to set the monitor object's total transfer size.  The method does not necessarily know the total size of the data transfer.  For example, a progress bar that tracks the progress of multiple data transfers would call the read_binary_values() method multiple times, so the method would not have visibility of the total transfer size.  Also, a monitor object that is not a progress bar might not care about total transfer size.  

However, a utility function that predicts the total data transfer size could be used by the caller while it is setting up the monitor object before beginning the data transfer.  This function could accept a number of parameters (header format, number of data points, data encoding, etc.) and return the total number of bytes required for the transaction.

Should I create a utility function as part of this PR?  If so, where should it go in the pyvisa namespace?

MatthieuDartiailh:
Sorry for the lack of response. This went out of my radar. A helper would make sense to me yes. You can add it under util. It will be by function dealing with binary message parsing which makes sense.

cwstryker:
@MatthieuDartiailh   If there is a willingness to merge this pull request, I will add a helper to handle header size calculation.  I should be able to work on that in July.  If not, I'll just close this PR.  Let me know.

MatthieuDartiailh:
Sorry for not following up on this. I am still willing to merge this, but I list track of it after my last message. Ping me when you manage to add the helper, and I will do my best to merge shortly.

cwstryker:
@MatthieuDartiailh  I believe the changes are complete.

pyvisa/util.py
```diff
@@ -1057,3 +1057,32 @@ def get_arch(filename: Union[str, Path]) -> List[ArchitectureType]:
             archs.append(ArchitectureType.AARCH64)
 
         return archs
+
+
+def message_size(
+    num_points: int,
+    point_size: Literal[1, 2] = 2,
```
MatthieuDartiailh, line 1064: Here people may need to use 4 or 8 when transferring floating points number. I wonder if proving the format of the data expected (b, h, f, d, etc) and doing the conversion internally would not be nicer from a user perspective.

pyvisa/util.py
```diff
@@ -1057,3 +1057,32 @@ def get_arch(filename: Union[str, Path]) -> List[ArchitectureType]:
...
+) -> int:
+    """Helper function to calculate a message size, including the header, in bytes.
+
+    Inputs:
+        num_points: Number of data points to transfer
+        point_size: Size of each data point in bytes
+        header_format: Specify "ieee" or "hp" or "empty" header format
+
+    Returns:
+        The total message size in bytes
```
MatthieuDartiailh, line 1075: The library use the Numpy-Style for docstrings for I would prefer to remain consistent
```suggestion
    Parameters
    -----------
    num_points : int
        Number of data points to transfer
    point_size : int
        Size of each data point in bytes
    header_format : str
        Specify "ieee" or "hp" or "empty" header format

    Returns
    --------
    int
        The total message size in bytes
```

CHANGES
```diff
@@ -18,9 +25,11 @@ PyVISA Changelog
     have been removed
   - `p.u.get_shared_library_arch` now returns `p.u.PEMachineType` instead of a string
   - `p.u.get_arch` now returns a list of `p.u.ArchitectureType`
-- update `Resource` context manager type annotation to preserve derived type PR #740
 - added support for using custom simulation files when using the `sim` backend
   with `pyvisa-shell`. PR #774
+- update `Resource` context manager type annotation to preserve derived type PR # 740
+- added support for an optional monitoring interface to the read_binary_values() and
+    query_binary_values() methods of message-based resources.
```
MatthieuDartiailh, line 32: Should be reverted.

MatthieuDartiailh (changes requested):
Thanks. I left a couple comments. I will try to address them myself if you cannot do so.

cwstryker:
> Thanks. I left a couple comments. I will try to address them myself if you cannot do so.

Thanks.  I'll work in this next weekend.

pyvisa/util.py
cwstryker: Done.  I changed the parameter from "point_size: int" to "datatype: BINARY_DATATYPES" and used the struct module to determine the point size from the BINARY_DATATYPES code.

pyvisa/util.py
cwstryker: Done/

CHANGES
cwstryker: Done (mostly).  Lines 31 and 32 above are still removed because the optional monitoring interface was not added to  version 1.14.0.

cwstryker:
@MatthieuDartiailh I think all of the issues have been resolved.

MatthieuDartiailh (approved):

MatthieuDartiailh:
Thanks for following up on this !

PR #762, opened 2023-07-21
Initial implementation of asyncio read by callback

tobinjones:
Hi, I'm submitting this draft pull request purely as a point of discussion. It is intended to be an update of the much older pull request #403 . I realise I've barged into this code base and made a lot of noise, so thank you for being kind about that. I don't want to ruffle any feathers.

@MatthieuDartiailh said he would like to see "a detailed plan (PEP-like)" about this, and this is my attempt at starting such a plan. If there is interest in what I have here so far then I would be happy to continue developing a more formal write-up.

In this discussion, there are two meanings of the word "async". I hope it's not too confusing.

1. In the back-end it refers to calling functions such as `viReadAsync`. From a python perspective these functions are normal blocking functions which happen to schedule something to be done, and will cause a VISA event to be raised when that thing is finished.
2. The proposal here to use `asyncio` python features to expose a high-level non-blocking API suitable for use in a larger async python program. Basically, python functions that start with `async def`

**Regarding the back-end:**

[As discussed in the previous PR](https://github.com/pyvisa/pyvisa/pull/403#issuecomment-1643662332), I believe that of the two available mechanisms for VISA async reads, that using the callback mechanism is superior for two main reasons:

1. It's conceptually nice to avoid polling or threading
2. It was the fastest method in my benchmark

**Regarding the front-end**

That leaves the question of the high-level async API you would like to expose. I'll need the maintainer's guidance on this since I am new to this code base and don't know your preferences and vision.

That said, I'll try and give some opinions and options.

If a user wants to use a function like `await inst.async_read()`, the resource must be set up first. We must install an event handler, set up some kind of synchronization method between the event handler and the asyncio code, and enable at least the `IO_COMPLETE` visa events. There is then associated teardown of those things.

I experimented with doing this setup and teardown on every single read, but unsurprisingly it was quite slow (it doubled the read time in my test). Therefore, I suggest that we offer a context manager to deal with this setup and teardown and require the async functions to be called from inside that context. I suggest something like an `async_io_context` method on `MessageBasedResource`. This has precedence, because it's similar to what you have already done with the `read_termination_context` method.

The big question remains then, where should the async methods be implemented?

Before we get into that, how many methods are we talking about? To begin with, the read* methods are probably the lowest-hanging fruit:

```
read_bytes
read_raw
read
read_ascii_value
read_binary_values
```

The `write*` and `query*` methods could also be made async but that strikes me as lower priority. Finally, it would be quite important to have a way to `await` a `VI_EVENT_SERVICE_REQ` event, for instruments which support that. So it's between 6 and 13 methods to implement.

So there is a decent amount of duplication here, but I think that's unavoidable. If we want a complete set of async methods then we must implement a complete set of async methods. Where possible, helper functions could be broken out to reduce duplication

So the question remains, where to put these methods? The two options I see for where to implement these methods are:

1. Directly on the existing MessageBasedResource. They would sit right beside their non-async partners, but raise an exception if not called from within the `async_io_context`. This would work something like:

```python
import pyvisa
import asyncio

async def main():
    rm = pyvisa.ResourceManager()
    inst = rm.open_resource('GPIB0::12::INSTR')
    with inst.async_io_context:
        inst.write("*IDN?")
        response = await inst.async_read_raw()
        print(response)

asyncio.run(main())
```

2. They could be implemented on another class, perhaps a subclass of `MessageBasedResource` called `AsyncMessageBasedResource`. This could be returned from the context manager like so:

```python
import pyvisa
import asyncio

async def main():
    rm = pyvisa.ResourceManager()
    inst = rm.open_resource('GPIB0::12::INSTR')
    with inst.async_io_context as ainst:
        assert(type(ainst) == AsyncMessageBasedResource)
        ainst.write("*IDN?") # Should this method be called on `inst` or `ainst`? Both work...
        response = await ainst.async_read_raw()
        print(response)

asyncio.run(main())
```

I think the former is simpler because it doesn't add anything to the class heirarchy and it doesn't require the user to deal with two "instrument" objects that are almost-but-not-quite-the-same.

However, it does mean that the "async_*" methods are there, hanging about, even for people who don't want them.

I would lean towards the first option unless there was a compelling reason not to. Thus, this is what I have implemented in the draft PR. I have only implemented `async_read_raw` for the moment. The other methods are a job for after some consensus!

Cheers, and I hope this is useful.

- [  ] The change is fully covered by automated unit tests
- [  ] Documented in docs/ as appropriate
- [  ] Added an entry to the CHANGES file

(status: draft)

### DIFF
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -7,15 +7,28 @@ This file is part of PyVISA.
 :license: MIT, see LICENSE for more details.
 
 """
+import asyncio
 import contextlib
 import struct
 import time
 import warnings
-from typing import Any, Callable, Iterable, Iterator, Optional, Sequence, Type, Union
+from typing import (
+    Any,
+    AsyncIterable,
+    Callable,
+    Iterable,
+    Iterator,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 from .. import attributes, constants, errors, logger, util
 from ..attributes import Attribute
 from ..highlevel import VisaLibraryBase
+from ..typing import VISAEventContext, VISAHandler, VISASession
 from .resource import Resource
 
 
@@ -75,6 +88,9 @@ class MessageBasedResource(Resource):
 
     #: Internal storage for the encoding
     _encoding: str = "ascii"
+
+    # Internal queue for async IO
+    _async_io_queue: Optional[asyncio.Queue] = None
 
     @property
     def encoding(self) -> str:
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Thanks for your work and sorry for taking so long to come back to you.

From my experience using `async`, the use of `async` tends to be pervasive. Basically when you use non-blocking operations you want as many things as possible to be non-blocking. To me this goes in the direction of having async resources which offers (only/mainly) non blocking operation and to keep enabled whatever machinery is needed as long as necessary and to instantiate them using a dedicated method on the resource manager. 

We could also offer functions to turn a sync resource into an async one even though I see no compelling use for it.

As you point out it means we need a parallel class hierarchy which has a cost but to me it appears more viable. WDYT ?

tobinjones:
Hey, no problem! I don't want to demand your time on this. You didn't ask me to but in with all this. Anyway, I've been pulled to different things for a bit, and I won't be back to working on this properly for a week or two. 

I agree with you that in an async application, you naturally want as much to be non-blocking as possible.

I think your preference for a parallel class heirarchy instantiated from the resource manager is probably fine. The two minor things I see are:
 - What to do with methods like `read_termination` and `write_termination`? I think they will be identical between the `MessageBasedResource` and `AsyncMessageBasedResource`. Do we just copy-paste them? Add another level to the class heirarchy? Factor them out into helper functions?
 - If the async stuff is separate like this it's all-or-nothing, so the whole feature has to go out at once.

I don't think either of those are big problems.

Just to follow up, I have two observations from what I've experimented with implementing this.

### 1. The callback->queue glue
The fundamental abstraction is the ability to view the VISA event callback API as a kind of async queue of events. That's where the "async-ness" comes from. All other methods and functionality (reading, querying, getting events such as service requests, etc) build on that core abstraction. So it's the part that needs the most thought and care up front. Currently I have implemented this with the [async_event_context](https://github.com/pyvisa/pyvisa/blob/57d38d181d9789fd0da028ee29210e0685fb4022/pyvisa/resources/messagebased.py#L654) method. There are several details still to be worked out here.
 
Firstly I'll point out that the VISAEventContext (`event_ctx`) handed to the callback is closed by the VISA library at the end of the callback. By the time a reader comes along and pulls an event off the async queue, the `event_ctx` is gone. Therefore the callback must read all the information required from the `event_ctx` before it puts the event on the queue.

The result of this is we can't just use `events.Event` objects on our async queue, because they require an active VISAEventContext to read from.

So if we can't put `events.Event` objects on the queue, what do we put? Plain dictionaries maybe? some sort of concrete event object that's like a simple dataclass? Then should they have anything to do with your Event heirarchy? I think this is a bit of a tricky question.

Finally, what do we expose as a queue API? Do we hand back a plain `asyncio.Queue`? Just a plain function that returns an awaitable coroutine? Or our own custom Queue object?

I suspect writing a very simple wrapper around asyncio.Queue might be a good idea. 

### 2. Concurrent code and atomicity
Async programming is closely related to concurrency. It also think it tends to make people a bit blasé about concurrency bugs. In particular I think there is a big question about atomicity.

First let me clarify that if someone uses an async resource in a basically sequential way - and the only concurrency is that there are maybe OTHER async resources running at the same time, or like a GUI or a web service or something, then there's basically no problem. Many people might use the library this way.

However, if someone was to take an async resource and do something like:
"""
results = await asyncio.gather(
 inst.query("Query1"),
 inst.query("Query2")
)
"""
Then we have some issues. At a low level - query is a write followed by a read. Then, each read might be chunked. We therefore have the issue of two things awaiting chunks, and there's no way to tell which `read` they belong to. This would lead to (at mininum) garbled responses, and probably errors due to not reading messages in their entirity. So I think there needs to be an `asyncio.Lock` around the lowest level `read_raw` method to avoid this. I've implemented that in the attached code.

I will point out that this doesn't guarantee that there will be no problems. I already said that a query is a `write` followed by a `read`. So, the code above will probably translate into:
"""
inst.write("Query1")
inst.write("Query2")
inst.read()
inst.read()
"""
... which would cause an error on most instruments I've used. So the question is -- do we implement locking so that each "query_*", "read_*", "write_*" command is guaranteed atomic? That requires more code and more care (for example you don't want `query` to acquire a lock and then call `read` which also tries to acquire the lock... that's a deadlock).

But this is not really unique to the `async` version. The plain old pyvisa would have the same problems if you tried to do parallel queries on an instrument from different threads. So I kind of lean towards only implementing the lock around the chunked reads since there's not really a downside to that, and just telling people "don't do parallel queries, your instrument won't like it". But I raise this as an issue for discussion. Cheers!

ddi-acassidy:
> However, it does mean that the "async_*" methods are there, hanging about, even for people who don't want them.

The [sniffio](https://pypi.org/project/sniffio/) module could be used to detect when an async library is in use. That way the normal read() function would still be blocking when used normally, but async if run within an asyncronous context.

MatthieuDartiailh:
the use of asgiref will come with a runtime cost that I am not exactly happy to force on all users. However feel free to make a POC if you are interested, personally I do not have the bandwidth to attempt to do so at the moment.

PR #811, opened 2024-03-27
Prevent duplicate paths from being added to the DLL paths

### LINKED ISSUE #807: Extra DLL paths from .pyvisarc are never removed from DLL search path
# Problem Description
When specifying extra DLL paths in the `.pyvisarc`, the DLL paths are added to the DLL search path each time a new `ResourceManager` is instantiated, but never removed. Because of this, at some point the DLL search path will become too long for windows to handle, which will result in a WinError 206.

# Steps to Reproduce
1. Have some extra DLL paths in the `.pyvisarc`, for example the Keysight IO libs:<br> `dll_extra_paths: C:\Program Files (x86)\Keysight\IO Libraries Suite\bin`
2. Open and close many `ResourceManager` instances after each other
```
import pyvisa

for i in range(1000):
    print(i)
    rm = pyvisa.ResourceManager()
    rm.close()
```
On my system, at `i = 297` this will result in
```
FileNotFoundError: [WinError 206] The filename or extension is too long:  'C:\\Program Files (x86)\\Keysight\\IO Libraries Suite\\bin'
```

# Possible Fix
In my opinion the issue is here:
https://github.com/pyvisa/pyvisa/blob/6bca99233dcf66742db54d6d410c9840efb18cb5/pyvisa/util.py#L150
The call to `os.add_dll_directory` returns a `os._AddedDllDirectory` object, which needs to be closed according to the python docs for `os`:
>"Remove the directory by calling close() on the returned object or using it in a [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement."
However, the  `os._AddedDllDirectory` is simply discarded here.

I am not yet sure about a fix, do we need to keep track of the `os._AddedDllDirectory` until the `ResourceManger` instance is closed?
Or could we return the path to the DLL library as a string, and later construct a `LibraryPath` object from the path in `get_library_paths`?

arr-ee:
This PR adds very rudimentary tracking of dll paths we already added in order to prevent exceeding path length limit if we create `ResourceManager`s in a loop.

- [x] Closes #807

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,9 +4,11 @@ PyVISA Changelog
 1.15.0 (unreleased)
 -------------------
 - added support for an optional monitoring interface object to the read_binary_values() and
-    query_binary_values() methods of message-based resources.
+  query_binary_values() methods of message-based resources.
 - added the message_size() helper function to calculate the estimated size of a data
-    transfer, including the header.
+  transfer, including the header.
+- prevent adding duplicate paths to the DLL search path during ResourceManager
+  instantiation #811
 
 1.14.1 (22-11-2023)
 -------------------
--- a/pyvisa/testsuite/test_util.py
+++ b/pyvisa/testsuite/test_util.py
@@ -18,6 +18,7 @@ from types import ModuleType
 from typing import Optional
 
 import pytest
+from pytest import LogCaptureFixture, MonkeyPatch
 
 from pyvisa import highlevel, util
 from pyvisa.ctwrapper import IVIVisaLibrary
@@ -45,7 +46,7 @@ class TestConfigFile(BaseTestCase):
             ]
         ):
             raise unittest.SkipTest(
-                ".pyvisarc file exists cannot properly test in this case"
+                ".pyvisarc file exists; cannot properly test in this case"
             )
         self.temp_dir = tempfile.TemporaryDirectory()
         os.makedirs(os.path.join(self.temp_dir.name, "share", "pyvisa"))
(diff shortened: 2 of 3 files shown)


arr-ee:
Some notes:

- I am not terribly thrilled with this solution, but, as mentioned in the original issue, without re-doing the way we approach config I don't see a significantly better way of tackling this.
- Not a huge fan of duplication in those tests; I don't mind moving them into a separate class and cleaning up at least some of the duplication, but didn't want to pollute the diff even more.
- Ditto for typehinting. I added it to the tests I touched, preempting mypy's complaints, but maybe it'd be better to move that change into a separate PR that fixes the entire file (or more)

arr-ee:
https://github.com/pyvisa/pyvisa/actions/runs/8428919122/job/23082317276#step:5:60 so, ruff has been failing for a minute now. Rapidly updating linter + no version constraints = randomly red CI, fun!

arr-ee:
https://github.com/pyvisa/pyvisa/pull/812 should hopefully resolve formatting issues (for now)

MatthieuDartiailh:
Can you rebase on main to check the changes to the CI configuration behaved as expected ?

arr-ee:
Will do

MatthieuDartiailh:
@arr-ee I took the liberty to rebase your changes. Hope you do not mind. Can you give this a quick look and I will merge ?

arr-ee:
Of course, thank you for picking this up.

Looks fine, I see the change that you made and it should be ok I imagine.

MatthieuDartiailh:
Thanks for having a look.

PR #816, opened 2024-04-08
add -v cmdline flag to pyvisa-shell for warn/info/debug logging output

### LINKED ISSUE #758: ThorLabs PM100USB not listed by pyvisa-shell
I have ThorLabs PM100USB meter that is correctly listed by `lsusb`:
```
Bus 001 Device 032: ID 1313:8072 ThorLabs PM100USB Power and Energy Meter Interface
```
How when I execute list command in `pyvisa-shell` no device is listed.
Does anyone know what might be wrong?

### LINKED ISSUE #791: [COM] Communication issue with Keysight DSOX4104A oscilloscope using USB
Instrument details
------------------
* Model: Keysight InfiniiVision DSOX 4104A Oscilloscope
* Communication: USB
* Link to the documentation (if available):

Output of `pyvisa-info`
-----------------------

```
Machine Details:
   Platform ID:    Linux-6.6.5-arch1-1-x86_64-with-glibc2.38
   Processor:

Python:
   Implementation: CPython
   Executable:     /home/ashish/pyvisa/bin/python
   Version:        3.11.6
   Compiler:       GCC 13.2.1 20230801
   Architecture:   ('x86', 64)
   Build:          Nov 14 2023 09:36:21 (#main)
   Unicode:        UCS4

PyVISA Version: 1.14.1

Backends:
   ivi:
      Version: 1.14.1 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.7.1
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR: Available via PyUSB (1.2.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.2.1). Backend: libusb1
      TCPIP INSTR: Available
         Resource discovery:
         - VXI-11: ok
         - hislip: ok
      TCPIP SOCKET: Available
      VICP INSTR:
         Please install PyVICP to use this resource type.
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
      GPIB INTFC:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
```

When trying to list the visa address of the attached instrument, I only get `ASRL<Num>::INSTR` addresses and not the oscilloscope. The interface of the oscilloscope lists a USB type address i.e. `USB0::2391::6048::MY61260137::INSTR`.
On running `lsusb`, I can confirm that the oscilloscope is connected to my laptop displaying the line:
```
Bus 003 Device 014: ID 0957:17a0 Agilent Technologies, Inc. DSO-X 4104A
```

jllanfranchi:
This commit adds `-v` (or `-vv` or `-vvv` or `-v -v` or `-v -v -v`) as command line flags that can be passed to `pyvisa-shell` such that WARN (`-v`), INFO (`-vv`), or DEBUG (`-vvv`) messages will be emitted when running the shell. This can help users understand why a device is not listed that they believe should be listed when using that tool. Guidance in documentation can be added (by me, in the near future) to tell users to utilize these options if things are not working correctly. As it stands now, `pyvisa-shell` just doesn't say anything, and the only way to get increased messages is to write a Python script to do this.

This commit, along with https://github.com/pyvisa/pyvisa-py/pull/423 in the pyvisa-py, (as soon as I also update documentation appropriately) will close #758 and I expect will also close #791. 

- [ ] Partially closes #758 and partially closes #791 (see also https://github.com/pyvisa/pyvisa-py/pull/423 (TODO: update documentation)
- [ ] The change is fully covered by automated unit tests (**Probably not necessary or possible for this?**)
- [ ] Documented in docs/ as appropriate (**TODO**)
- [ ] Added an entry to the CHANGES file (**TODO**)

### DIFF
--- a/pyvisa/cmd_line_tools.py
+++ b/pyvisa/cmd_line_tools.py
@@ -40,7 +40,7 @@ def visa_main(command: Optional[str] = None) -> None:
         dest="verbosity",
         action="count",
         default=0,
-        help="verbosity; repeat up to 3 times for increased output",
+        help="verbosity; use up to 3 times for increased output",
     )
 
     if not command:
@@ -53,7 +53,7 @@ def visa_main(command: Optional[str] = None) -> None:
     args = parser.parse_args()
     logging.basicConfig(
         level={0: logging.ERROR, 1: logging.WARN, 2: logging.INFO, 3: logging.DEBUG}[
-            args.verbosity
+            min(3, args.verbosity)
         ]
     )
     if command:
--- a/pyvisa/cmd_line_tools.py
+++ b/pyvisa/cmd_line_tools.py
@@ -52,9 +52,7 @@ def visa_main(command: Optional[str] = None) -> None:
 
     args = parser.parse_args()
     logging.basicConfig(
-        level=(logging.WARN, logging.INFO, logging.DEBUG)[
-            min(2, args.verbosity)
-        ]
+        level=(logging.WARN, logging.INFO, logging.DEBUG)[min(2, args.verbosity)]
     )
     if command:
         args.command = command
--- a/pyvisa/cmd_line_tools.py
+++ b/pyvisa/cmd_line_tools.py
@@ -52,8 +52,8 @@ def visa_main(command: Optional[str] = None) -> None:
 
     args = parser.parse_args()
     logging.basicConfig(
-        level={0: logging.ERROR, 1: logging.WARN, 2: logging.INFO, 3: logging.DEBUG}[
-            min(3, args.verbosity)
+        level=(logging.WARN, logging.INFO, logging.DEBUG)[
+            min(2, args.verbosity)
         ]
     )
     if command:
--- a/pyvisa/cmd_line_tools.py
+++ b/pyvisa/cmd_line_tools.py
@@ -9,6 +9,7 @@ This file is part of PyVISA.
 """
 
 import argparse
+import logging
 from typing import Optional
 
 
@@ -34,6 +35,13 @@ def visa_main(command: Optional[str] = None) -> None:
         default=None,
         help="backend to be used (default: ivi)",
     )
+    parser.add_argument(
+        "-v",
+        dest="verbosity",
+        action="count",
+        default=0,
+        help="verbosity; repeat up to 3 times for increased output",
+    )
 
     if not command:
         subparsers = parser.add_subparsers(title="command", dest="command")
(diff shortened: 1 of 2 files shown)


pyvisa/cmd_line_tools.py
```diff
@@ -43,6 +51,11 @@ def visa_main(command: Optional[str] = None) -> None:
         subparsers.add_parser("shell", help="start the PyVISA console")
 
     args = parser.parse_args()
+    logging.basicConfig(
+        level={0: logging.ERROR, 1: logging.WARN, 2: logging.INFO, 3: logging.DEBUG}[
```
arr-ee, line 55: any reason for an int-indexed dict instead of a tuple/list?

arr-ee:
Good call on making logging easier to enable here.

I think that `WARN` might be a more useful default value since this is an interactive session — wdyt?

pyvisa/cmd_line_tools.py
jllanfranchi: None at all! A tuple or list would be more compact, no less explicit, & works just as well.

jllanfranchi:
One issue with making WARN default is that it changes the current behavior. If we're willing to change the current behavior, then I'm fine with that.

Another consideration is whether we want to add a command line option that _decreases_ the verbosity level back to ERROR (presumably the default now), if users don't want to see WARN-level messages at all? (Seems like WARN level messages shouldn't be _too_ annoying, though...)

arr-ee:
> and would kinda prefer it, so users get useful advice without having to figure out there's a -v option in the first place

This is exactly why I think `WARN` should be the default, yes. I am not too bothered about change of behaviour — although I'd love to hear from folks using the shell if I'm wrong!

> Another consideration is whether we want to add a command line option that decreases the verbosity level back to ERROR 

This is fair, `-q/--quiet`?

pyvisa/cmd_line_tools.py
MatthieuDartiailh: Can you address this then ?

MatthieuDartiailh (changes requested):
I agree WARN is a sane default for such a case. We can `-q` but I do not see people needing it to be honest. If anybody is using the shell enough to be bothered by a warning they may be better served by a script.

MatthieuDartiailh:
ping  @jllanfranchi could you address the small change request and I will merge.

MatthieuDartiailh:
I took the liberty to rebase your PR and make the small changes we discussed. I will merge by the end of the week.

PR #859, opened 2024-10-01
cli: unshare pyvisa cli command implementation

henriquesimoes:
pyvisa-shell and pyvisa-info have nothing in common besides being PyVISA CLI commands. Notice that all options are currently targeted for pyvisa-shell [1], even though argparse happily reports `--backend` and `-v` options for pyvisa-info as well. In addition to that, everything else in the shared `pyvisa_main()` is conditional on the given `command`.

Implement them separately, including the function body and argument handling. This effectively reduces the amount of code and complexity for having them available. It also enables proper handling of the commands documentation, including description line. If arguments ever needs to be shared by both commands, there can be a shared function that equally populate their parsers.

Parsers are created by standalone functions, so that other tools can easily use them, such as argparse-manpage. Even though pyvisa-info requires no extra argument, we still define a parser for it, so that users can still provide `--help` to get further instructions.

Since 2095ebae1ca9 ("remove deprecated feature planned for removal in 1.12"), there is also no `__main__` entry point for module commands, which renders subcommand parsing also not needed. Testing argument handling can therefore be also removed.

[1] 4af61c172492 ("add -v cmdline flag to pyvisa-shell for warn/info/debug logging output") intentionally attempts to add an option specifically for pyvisa-shell, but also adds it for pyvisa-info.

- [x] Executed `ruff check . && ruff format --check` with no errors

I haven't added tests to further exercise the code paths for options, as 4af61c172492 itself does not add it for verbose levels, and testing backends is not straightforward. Let me know if you think that is a required step to merge this.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@ PyVISA Changelog
 
 1.15.0 (unreleased)
 -------------------
+- make `pyvisa-shell` and `pyvisa-info` handle arguments independently
 - add support for Python 3.13 PR #852
 - drop support for Python 3.8 and 3.9 PR #852
 - add a -v option to pyvisa-shell to increase log verbosity and set default to WARN PR #816
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,7 @@ PyVISA Changelog
 
 1.15.0 (unreleased)
 -------------------
-- make `pyvisa-shell` and `pyvisa-info` handle arguments independently
+- make `pyvisa-shell` and `pyvisa-info` handle arguments independently PR #859
 - add support for Python 3.13 PR #852
 - drop support for Python 3.8 and 3.9 PR #852
 - add a -v option to pyvisa-shell to increase log verbosity and set default to WARN PR #816
--- a/pyvisa/cmd_line_tools.py
+++ b/pyvisa/cmd_line_tools.py
@@ -35,7 +35,7 @@ def get_shell_parser() -> argparse.ArgumentParser:
         dest="verbosity",
         action="count",
         default=0,
-        help="verbosity; use up to 3 times for increased output",
+        help="verbosity; use up to 2 times for increased output",
     )
 
     return parser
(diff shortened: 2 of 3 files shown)


pyvisa/cmd_line_tools.py
```diff
@@ -10,22 +10,17 @@
...
-    command : str, optional
-        What command to invoke, if None the value is read from the command
-        line arguments
+def get_shell_parser() -> argparse.ArgumentParser:
+    """Create argument parser for pyvisa-shell program."""
 
+    description = """
+    PyVISA interactive console for debugging and testing with PyVISA-compatible
+    devices.
     """
```
henriquesimoes, line 21: I'm not myself familiar enough with the tool to provide a concise description. Any suggestions on the topic are very welcome.

pyvisa/cmd_line_tools.py
```diff
@@ -86,9 +56,31 @@ def _create_backend_str(args: argparse.Namespace) -> str:
...
+        level=(logging.WARN, logging.INFO, logging.DEBUG)[min(2, args.verbosity)]
+    )
+
+    backend = _create_backend_str(args)
+    shell.main(backend)
+
+
+def get_info_parser() -> argparse.ArgumentParser:
+    """Create argument parser for pyvisa-info program."""
+    parser = argparse.ArgumentParser(description="PyVISA diagnostic information tool.")
```
henriquesimoes, line 74: Same thing here.

CHANGES
```diff
@@ -3,6 +3,7 @@ PyVISA Changelog
 
 1.15.0 (unreleased)
 -------------------
+- make `pyvisa-shell` and `pyvisa-info` handle arguments independently
```
MatthieuDartiailh, line 6: Please add the PR number

MatthieuDartiailh (commented):
LGTM. I will test locally, add the PR number to the release notes and merge. Thanks for your contribution.

PR #866, opened 2024-10-31
Set 4.0.0 as the lower bound for `typing_extensions`

### LINKED ISSUE #865: `typing_extension` does not have a lower bound
`rname.py` [has this import](https://github.com/pyvisa/pyvisa/blob/a422da144ba054c76f282589330570beca872c6a/pyvisa/rname.py#L26):

```python
from typing_extensions import ClassVar, Self
```

`Self` was added to `typing_extensions` [in 4.0.0](https://github.com/python/typing_extensions/blob/67c16e190e7ebe6a9130b63839520ce01bdd6422/CHANGELOG.md#added-in-version-400). However, [as it is currently declared](https://github.com/pyvisa/pyvisa/blob/a422da144ba054c76f282589330570beca872c6a/pyproject.toml#L31), `typing_extensions` has no lower bound:

```toml
dependencies = ["typing_extensions"]
```

This caused [some confusion](https://stackoverflow.com/q/79146179).

InSyncWithFoo:
Resolves #865.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,7 @@ PyVISA Changelog
   transfer, including the header.
 - prevent adding duplicate paths to the DLL search path during ResourceManager
   instantiation #811
+- set 4.0.0 as the lower bound for `typing_extensions` PR #866
 
 1.14.1 (22-11-2023)
 -------------------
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh (approved):

MatthieuDartiailh:
Thanks

PR #874, opened 2024-11-30
Use fewest transfers possible in `read_binary_values`

aWZHY0yQH81uOYvH:
See https://github.com/pyvisa/pyvisa-py/issues/472 for more details. Essentially, I have an instrument that will lose data when too many small reads are issued. This PR makes it so `read_binary_values`, when not passed an explicit `chunk_size`, uses one transfer of default length to read the data header, then if there is more data to read, it reads it all in one more transfer. This should be more efficient and solves a problem with my instrument. Behavior for users that specify `chunk_size` is unaffected.

In order for this to work with the `pyvisa-py` backend, https://github.com/pyvisa/pyvisa-py/pull/470 needs to be merged since it currently further breaks up reads.

- [x] Documented in docs/as appropriate

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,8 @@ PyVISA Changelog
 - add a -v option to pyvisa-shell to increase log verbosity and set default to WARN PR #816
 - added support for an optional monitoring interface object to the read_binary_values() and
   query_binary_values() methods of message-based resources.
+- automatically determine default chunk_size in read_binary_values() so only one or two
+  transfers are used for all data PR #874
 - added the message_size() helper function to calculate the estimated size of a data
   transfer, including the header.
 - prevent adding duplicate paths to the DLL search path during ResourceManager
--- a/docs/source/introduction/rvalues.rst
+++ b/docs/source/introduction/rvalues.rst
@@ -134,10 +134,12 @@ An implementation of a progress bar might look like:
 If you can read without any problem from your instrument, but cannot retrieve
 the full message when using this method (VI_ERROR_CONN_LOST,
 VI_ERROR_INV_SETUP, or Python simply crashes), try passing different values for
-``chunk_size`` (the default is 20*1024). The underlying mechanism for this issue
-is not clear but changing ``chunk_size`` has been used to work around it. Note
-that using  larger chunk sizes for large transfer may result in a speed up of
-the transfer.
+``chunk_size`` (the default is 20*1024 for the first transfer to get the header,
+then the remaining data length in a second transfer if necessary). The underlying
+mechanism for this issue is not clear but changing ``chunk_size`` has been used to
+work around it. Note that using larger chunk sizes for large transfer may result in
+a speed up of the transfer, which is why the default is to use the largest
+possible ``chunk_size`` once the data length is known.
 
 In some cases, the instrument may use a protocol that does not indicate how
 many bytes will be transferred. The Keithley 2000 for example always returns the
--- a/pyvisa/resources/messagebased.py
+++ b/pyvisa/resources/messagebased.py
@@ -598,7 +598,9 @@ class MessageBasedResource(Resource):
              number of bytes based on the datatype. Defaults to 0.
         chunk_size : int, optional
             Size of the chunks to read from the device. Using larger chunks may
-            be faster for large amount of data.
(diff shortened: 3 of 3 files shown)


CHANGES
```diff
@@ -9,6 +9,8 @@ PyVISA Changelog
 - add a -v option to pyvisa-shell to increase log verbosity and set default to WARN PR #816
 - added support for an optional monitoring interface object to the read_binary_values() and
   query_binary_values() methods of message-based resources.
+- automatically determine default chunk_size in read_binary_values() so only one or two
+  transfers are used for all data
```
MatthieuDartiailh, line 13: Please add the PR number.

MatthieuDartiailh (commented):
LGTM.

Could you update the release notes with the PR number ? Also could you add a note in the docs about the new behavior ?

aWZHY0yQH81uOYvH:
Sorry, didn't notice the other items in the changelog had PR numbers. I did update the docstring for the `read_binary_values` function to describe the new default behavior, which shows up in the docs. Would you like it mentioned somewhere else also?

MatthieuDartiailh:
You need to adjust the section `Reading binary values` in docs/source/introduction/rvalues.rst

aWZHY0yQH81uOYvH:
Done.

MatthieuDartiailh:
Thanks, it looks good. I am wondering if this could break when transferring very long responses in which case it may be better to cap the maximum chunck_size. I will try to test either tomorrow or next week before merging.

aWZHY0yQH81uOYvH:
I wonder that as well.

Looks like the length field in the USBTMC header is a 4-byte value (up to 4GiB), while the IEEE header represents the length with up to 9 decimal digits (up to 999MB), so it should be impossible to overflow the length field of the USB header. Similarly, looks like the HP header is a 2-byte length value, so that should also be fine.

I don't have other equipment to test this with, so further testing would be great to see if there are issues below these limits.

MatthieuDartiailh:
Sorry I still did not get a chance to test this. If anybody has the bandwidth to do it it would be welcome.

aWZHY0yQH81uOYvH:
I just tested this with a Keysight DSOX4024A, Keithley 2450, Tektronix AFG 31000, and Keithley 2230G. All instruments seemed to work as expected with a few basic commands, and I was able to transfer 2MB of data from the scope without issue about 20% faster than without this PR (in 100ms). My tests aren't extensive, but I didn't have any issues.

MatthieuDartiailh:
Great, that's already a very good data point. With your DSO, can you try transferring several 100 MB ? If it succeeds, I think we should be good to go

aWZHY0yQH81uOYvH:
I had a look through the programming manual and it seems I misread that the maximum transfer is actually 10M 10-bit points (limited to 4M I think on this model), not 1M. I can try that next time I go to the lab.

Even now that I have access to a lab with much nicer equipment, I still don't think I have a device that can do several 100MB all in one go.

MatthieuDartiailh:
So I finally managed to run some test transferring 200M (after transferring easily 100k), but I ran out of time before figuring out what I was doing wrong (or not transferring so much data is slow) (and crashed the instrument through random disconnection). I will try again later this week.

MatthieuDartiailh:
Ok, I can finally report some interesting testing results.
I tried transferring both 60k points and 100M points over TCPIP::SOCKET using both the ivi backend (Keysight VISA) and pyvisa-py on Windows.

Using the Pyvisa-py backend this PR and the main branch performs equally (within the error margin associated with such large trasnfers).
Using the IVI backend with Keysight VISA things started to go sideways. Long story short when the chunk_size is larger than 2**18 (at least 2**19 is too much) the transfers slows to a crawl (I did not wait for it to finish but using tdqm I could see it was still doing something).

As a consequence I think we should introduce a chunk_size_limit at the resource level to prevent such slow down to occur. The default value could be 2**18 (i.e. 256k). This way we should get the best of both worlds. 

WDYT @aWZHY0yQH81uOYvH ? If you agree let me know if you want to do the changes yourself or if I can take care of them.

MatthieuDartiailh:
Another option could be to have the default value defined on the library class and put no limit with pyvisa-py.

PR #878, opened 2024-12-09
attributes: add VICP resources to supported attributes.

### LINKED ISSUE #876: Cannot set read_termination property on VICP instruments (LeCroy)
When attempting to set the `read_termination` property on instruments connected with VICP (LeCroy scopes), a `VI_ERROR_NSUP_ATTR` error is raised saying the attribute is not defined or supported. However `hasattr` returns `True` and the attribute can be read (default value is `None`), it just cannot be set to a new value.

```
>>> import pyvisa
>>> scope = pyvisa.ResourceManager('@py').open_resource('VICP::10.33.90.85::INSTR')
>>> print(scope.read_termination)
None
>>> scope.query('*IDN?')
'LECROY,WR64XI-A,LCRY0617N55856,8.1.2\n'
>>> scope.read_termination = '\n'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\seth.pevsner\AppData\Roaming\Python\Python311-32\site-packages\pyvisa\resources\messagebased.py", line 109, in read_termination
    self.set_visa_attribute(
  File "C:\Users\seth.pevsner\AppData\Roaming\Python\Python311-32\site-packages\pyvisa\resources\resource.py", line 378, in set_visa_attribute
    return self.visalib.set_attribute(self.session, name, state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\seth.pevsner\AppData\Roaming\Python\Python311-32\site-packages\pyvisa_py\highlevel.py", line 659, in set_attribute  
    return self.handle_return_value(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\seth.pevsner\AppData\Roaming\Python\Python311-32\site-packages\pyvisa\highlevel.py", line 251, in handle_return_value
    raise errors.VisaIOError(rv)
pyvisa.errors.VisaIOError: VI_ERROR_NSUP_ATTR (-1073807331): The specified attribute is not defined or supported by the referenced object.
>>> hasattr(scope, 'read_termination')
True
```

This appears to be due to the `AttrVI_ATTR_TERMCHAR` and `AttrVI_ATTR_TERMCHAR_EN` classes in `attribute.py` not having VICP listed as a supported interface type for this attribute. Adding the following line on both of them fixes the issue:

image

I think the `AttrVI_ATTR_SEND_END_EN` should also have this line. The `TCPIPInstrVicp` in `tcpip.py` in `pyvisa-py` also assigns a value to this attribute so I assume it also should be supported on VICP but that the same problem would occur if attempt is made to set its value, though I'm not sure which property on the resource is linked to that attribute.

Anyway, with the fix applied for `read_termination`, everything works.

```
>>> import pyvisa
>>> scope = pyvisa.ResourceManager('@py').open_resource('VICP::10.33.90.85::INSTR')
>>> print(scope.read_termination)
None
>>> scope.query('*IDN?')
'LECROY,WR64XI-A,LCRY0617N55856,8.1.2\n'
>>> scope.read_termination = '\n'
>>> scope.query('*IDN?')
'LECROY,WR64XI-A,LCRY0617N55856,8.1.2'
```

MatthieuDartiailh:
Since I never found a VICP spec I went with what pyvisa-py supports plus some that are supported by both TCPIP INSTR and SOCKET.

- [x] Closes #876

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,7 @@ PyVISA Changelog
 - prevent adding duplicate paths to the DLL search path during ResourceManager
   instantiation #811
 - set 4.0.0 as the lower bound for `typing_extensions` PR #866
+- list VICP resources in the appropriates attributes PR #878
 
 1.14.1 (22-11-2023)
 -------------------
--- a/pyvisa/attributes.py
+++ b/pyvisa/attributes.py
@@ -761,6 +761,7 @@ class AttrVI_ATTR_SEND_END_EN(BooleanAttribute):
         (constants.InterfaceType.gpib, "INTFC"),
         (constants.InterfaceType.tcpip, "INSTR"),
         (constants.InterfaceType.tcpip, "SOCKET"),
+        (constants.InterfaceType.vicp, "INSTR"),
         (constants.InterfaceType.usb, "INSTR"),
         (constants.InterfaceType.usb, "RAW"),
         (constants.InterfaceType.vxi, "INSTR"),
@@ -790,6 +791,7 @@ class AttrVI_ATTR_SUPPRESS_END_EN(BooleanAttribute):
         (constants.InterfaceType.gpib, "INSTR"),
         (constants.InterfaceType.tcpip, "INSTR"),
         (constants.InterfaceType.tcpip, "SOCKET"),
+        (constants.InterfaceType.vicp, "INSTR"),
         (constants.InterfaceType.usb, "INSTR"),
         (constants.InterfaceType.usb, "RAW"),
         (constants.InterfaceType.vxi, "INSTR"),
@@ -823,6 +825,7 @@ class AttrVI_ATTR_TERMCHAR_EN(BooleanAttribute):
         (constants.InterfaceType.asrl, "INSTR"),
         (constants.InterfaceType.tcpip, "INSTR"),
         (constants.InterfaceType.tcpip, "SOCKET"),
+        (constants.InterfaceType.vicp, "INSTR"),
         (constants.InterfaceType.usb, "INSTR"),
         (constants.InterfaceType.usb, "RAW"),
         (constants.InterfaceType.vxi, "INSTR"),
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
@spevs Can you check it does fix your issue ? Also since I do not have access to a VICP instrument can you quickly the other attributes ?

spevs:
> @spevs Can you check it does fix your issue ? Also since I do not have access to a VICP instrument can you quickly the other attributes ?

I've checked the `vicp-attributes` branch does fix my issue.

I checked the other attributes that you added VICP to in the commit by reading each one with `get_visa_attribute()`; not sure if this is the correct approach (I'm not really a software developer) but at least `get_visa_attribute(pyvisa.constants.VI_ATTR_TERMCHAR)` gives an error on main but works in the vicp branch, so it seems like it's the right thing to do...

For the ones you added a line to include VICP instruments, most of them worked, but a few still failed:
* `VI_ATTR_RD_BUF_SIZE`
* `VI_ATTR_WR_BUF_SIZE`
* `VI_ATTR_TCPIP_KEEPALIVE`

If I step through the code, at the time I'm checking the attributes indeed the buffer size attributes don't seem to exist on the object. However, `tcpip_keepalive` does exist, at the bottom of the `attrs` list. But still gives the unsupported error.

image

When I step into `sess.get_attribute` and then into `in_resource`, this is the check that fails. On the left side I can see VICP is not present in the supported resources list for the keepalive attribute, only TCPIP socket and instrument.

image

However, if I scroll further down in the `attributes.py` file to the definition, I can see the source code does have VICP defined for the keepalive attribute:

image

So it seems like it's being changed somewhere at runtime... I'm not too familiar with how this code works so I'm not sure what to make of this.

MatthieuDartiailh:
Thanks for testing. Indeed pyvisa-py patches the keep alive attribute to expose it on VXI-11 while according to the standard it should not exist. I will make a fix there too. For the other two they are not yet supported on pyvisa-py but it is fine to error.

PR #913, opened 2025-05-22
Fix handling of missing secondary address for GPIB resources

purnimab:
A GPIB instrument without a secondary address will give a default value of `VI_NO_SEC_ADDR = 0xFFFF` > 160, which then throws an `OverflowError: int too big to convert`. This patch modifies `GPIBCommand.secondary_address` to check for this default value, rather than assuming the default value is 0.

Although I did not submit an issue for this, this fixes a bug in which calling `GPIBInterface.group_execute_trigger` to trigger an instrument without a secondary address results in the above error.

- [ ] Added an entry to the CHANGES file


MatthieuDartiailh:
Thanks

PR #914, opened 2025-05-23
Add support for extended block lengths

### LINKED ISSUE #911: Support for extended block lengths
The IEEE 488.2 definite-length block format specifies the length of the block (number of bytes) with a decimal number in ASCII. This is preceded by a single decimal digit (0-9) which indicates how many digits long the length is. This means blocks cannot be 1 GB or larger, since the length indicator is limited to 9 decimal digits and hence the maximum representable value is 999,999,999 bytes.

As high end equipment has now reached these levels of memory, manufacturers have extended this system in various ways.

LeCroy and Tektronix have extended the length digit into hexadecimal, so the maximum length is now 15 digits (allowing up to 1 PB). Rohde & Schwarz have taken a different approach; they remove the initial length digit, and instead delimit the block length with parentheses, when the block length reaches 1 GB or larger. So for a 1,234,567,890 byte block (10-digit length) in these two systems, the header would be:

- `#A1234567890DATA` (Tektronix / LeCroy)
- `#(1234567890)DATA` (Rohde & Schwarz)

(At the moment it seems Keysight does not support ≥1 GB block lengths at all yet, according to the latest version manuals on their website, but I suspect they would follow the same path as Tek and LeCroy if they ever add support for it).

The Tek/LeCroy hexadecimal extension is supported easily enough by simply changing `int(header_length)` to `int(header_length, base = 16)` when evaluating the length digit, all previously valid inputs (using only decimal digits 0-9) will have the same value when interpreted in hex.

The R&S system requires some more work. I have written an implementation available in a fork here, modifying `parse_ieee_block_header`: https://github.com/spevs/pyvisa/blob/extended-block-length/pyvisa/util.py

I have tested this with a R&S scope with 1 GSa of memory and it works. I also tested that normal-sized block reads from various other instruments (not using any extended length systems) still work as expected. Unfortunately I do not have access to any Tek or LeCroy scopes with large enough sample memory that the extended length hexadecimal digit would come into play. But I am quite sure the `base = 16` would do the trick, that one seems hard to go wrong.

So far I have not covered block writes, which work the same way for the respective manufacturers. Right now this would require some user interface change as the user would need some way to specify which extended length syntax to use, unless it were to be auto-detected based on manufacturer in the IDN (which feels quite hacky). Ideally R&S should add support for accepting writes in the hexadecimal length format in addition to their parentheses format, which I will ask them to do. Most likely they would keep reads the same in order not to break existing code, but if they can at least accept writes in either format, it would eliminate the need for the user to have to specify the format, as the hex extension could be used for all writes, and reads can be auto-detected.

spevs:
This PR adds the following:

* `parse_ieee_block_header` now supports decoding blocks ≥ 1 GB using hexadecimal length digit (Tektronix/LeCroy)
* New function `parse_rs_block_header` supports decoding blocks ≥ 1 GB using parentheses-delimited length (Rohde & Schwarz)
* New `header_fmt = "rs"` option added to `read/query_binary_values()`. Internally this first calls the function `parse_ieee_or_rs_block_header`, which detects whether the block length begins with a parentheses and then calls the `ieee` or `rs` parser as appropriate. This implementation is used because R&S only uses this format when the block length exceeds 1 GB, so it's best to use automatic detection rather than need to have the user explicitly know whether it is using the parentheses format or not
* Exposes the `length_before_block` and `raise_on_late_block` options to the user interface of `read/query_binary_values()`, to allow user to set these values. Previously these were not accessible by the user, and always set to 25 and False. Some Tektronix scopes can return trace metadata in ASCII in front of the block data (before the block header) so a late block header >25 bytes can be expected, but previously it will always raise a warning :
```
UserWarning: The beginning of the block has been found at 156 which is an unexpectedly large value. The actual block may have been missing a beginning marker but the block contained one:
bytearray(b'1;8;BINARY;RI;MSB;"Ch1, DC coupling, 1.000V/div, 200.0us/div, 400000 points, Sample mode";1000;Y;"s";5.0000E-9;0.0000;124880;"V";40.0000E-3;1.0000;0.0000;1;#41000@AA?@?A?@@AB?A@C@A@')
```
In future it will be useful to add some way of retrieving this message in front of the block header (currently it is thrown away) but that is not included in this PR.

As a note, since the `header_fmt = rs` option in this implementation will interpret both IEEE (including the hex digit extension) and RS formatted blocks, it is somewhat "universal" for definite-length blocks, but doesn't detect indefinite length blocks if they happen to start with `"("`. The R&S format is only used for definite length blocks so I didn't think this would be a concern. But the user should be careful not to specify `header_fmt = rs` if an indefinite block is expected.

For the hex extension, I added a check to make sure the block length makes sense if the digit is 0xA to detect if an HP block is being decoded with the IEEE parser (the block length should be much more than 64K if there are 0xA = 10 digits, if it's not then it must have been an HP block, and an exception is raised with a descriptive message). But I don't have a way of testing this since I don't have any instruments that return blocks in HP format.

I tested that the R&S block decoding works, and the regular IEEE block decoding also still works.

- [x] Closes #911

### DIFF
--- a/pyvisa/testsuite/test_util.py
+++ b/pyvisa/testsuite/test_util.py
@@ -265,7 +265,7 @@ class TestParser(BaseTestCase):
 
         # Test handling definite length block
         p = util.from_ieee_block(
-            b"#214" + s[2:],
+            b"#256" + s[2:],
             datatype="f",
             is_big_endian=False,
         )
@@ -282,7 +282,7 @@ class TestParser(BaseTestCase):
 
         # Test handling definite length block with HP header format
         p = util.from_hp_block(
-            b"#A\x0e\x00" + s[2:],
+            b"#A\x38\x00" + s[2:],
             datatype="f",
             is_big_endian=False,
             container=partial(array.array, "f"),
@@ -292,7 +292,7 @@ class TestParser(BaseTestCase):
 
         # Test handling definite length block with R&S header format
         p = util.from_rs_block(
-            b"#(14)" + s[2:],
+            b"#(56)" + s[2:],
             datatype="f",
             is_big_endian=False,
         )
@@ -302,7 +302,7 @@ class TestParser(BaseTestCase):
         # Test handling definite length block with IEEE header format with
         # the automatic format determination
         p = util.from_ieee_or_rs_block(
-            b"#214" + s[2:],
+            b"#256" + s[2:],
             datatype="f",
             is_big_endian=False,
         )
@@ -312,7 +312,7 @@ class TestParser(BaseTestCase):
         # Test handling definite length block with R&S header format with
         # the automatic format determination
         p = util.from_ieee_or_rs_block(
-            b"#(214)" + s[2:],
+            b"#(56)" + s[2:],
             datatype="f",
             is_big_endian=False,
         )
@@ -333,7 +333,7 @@ class TestParser(BaseTestCase):
         # automatic format determination
         with pytest.raises(ValueError):
             p = util.from_ieee_or_rs_block(
-                b"#A\x0e\x00" + s[2:],
+                b"#A\x38\x00" + s[2:],
                 datatype="f",
                 is_big_endian=False,
             )
@@ -471,9 +471,9 @@ class TestParser(BaseTestCase):
     def test_malformed_rs_binary_block_header_no_closing_parenthesis(self):
         values = list(range(10))
         for header, tb, fb in zip(
-            ("rs"),
-            (util.to_rs_block),
-            (util.from_ieee_or_rs_block),
+            ("rs",),
+            (util.to_rs_block,),
+            (util.from_ieee_or_rs_block,),
         ):
             block = tb(values, "h", False)
             index = block.find(b")")
@@ -484,9 +484,9 @@ class TestParser(BaseTestCase):
     def test_malformed_rs_binary_block_header_late_closing_parenthesis(self):
         values = list(range(100))
         for header, tb, fb in zip(
-            ("rs"),
-            (util.to_rs_block),
-            (util.from_ieee_or_rs_block),
+            ("rs",),
+            (util.to_rs_block,),
+            (util.from_ieee_or_rs_block,),
         ):
             block = tb(values, "h", False)
             index = block.find(b")")
@@ -536,7 +536,7 @@ class TestParser(BaseTestCase):
             if header == "ieee":
                 parse = util.parse_ieee_block_header
             elif header == "hp":
-                partial(util.parse_hp_block_header, is_big_endian=False)
+                parse = partial(util.parse_hp_block_header, is_big_endian=False)
             elif header == "rs":
                 parse = util.parse_rs_block_header
             elif header == "ieee_or_rs":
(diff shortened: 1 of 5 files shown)


spevs:
I've added support for writes in the R&S format using `header_fmt = 'rs'`.

MatthieuDartiailh:
Perfect ! Could you just add a changelog entry and I will merge (sorry for not mentioning it earlier).

MatthieuDartiailh:
Actually Codecov just reminded that for the pieces of code under util we could/should have some tests (under pyvisa/testsuite).

spevs:
Also just to put it in writing, I also realized a correction about my original note:

> As a note, since the header_fmt = rs option in this implementation will interpret both IEEE (including the hex digit extension) and RS formatted blocks, it is somewhat "universal" for definite-length blocks, but doesn't detect indefinite length blocks if they happen to start with "(". The R&S format is only used for definite length blocks so I didn't think this would be a concern. But the user should be careful not to specify header_fmt = rs if an indefinite block is expected.

Actually it will handle indefinite block lengths too since they always start with a zero, not just the data straight away. So they will always be detected as "not starting with parentheses" and branch into the normal IEEE decoder and be handled there. So the "rs" header format option (for reads) should be able to actually handle anything except HP formatted blocks.

MatthieuDartiailh:
There are still some uncovered error paths. If you cannot add the relevant tests, I will have a look at it next week.

spevs:
Thanks. I added a test case for 537, although in normal circumstances I realized this is not reachable, because this function wouldn't be called directly, it would be called by `parse_ieee_or_rs_block_header`, but that function does its own check of the same in order to determine which way to branch. So the check is always guaranteed to pass in the child function.

Line 806 check is unneeded, since R&S format doesn't have the concept of an indefinite block length. I've removed it.

Line 809 I'm not sure why it's missed, I think it should have been covered already under `test_binary_block_shorter_than_advertized`. Anyway I added another test case to try to force it to reach.

pyvisa/testsuite/test_util.py
```diff
@@ -383,62 +444,106 @@ def fblock(block, cont):
     def test_bytes_binary_block(self):
         values = b"dbslbw cj saj \x00\x76"
         for block, tb, fb in zip(
-            ("ieee", "hp"),
-            (util.to_ieee_block, util.to_hp_block),
-            (util.from_ieee_block, util.from_hp_block),
+            ("ieee", "hp", "rs"),
+            (util.to_ieee_block, util.to_hp_block, util.to_rs_block),
+            (util.from_ieee_block, util.from_hp_block, util.from_ieee_or_rs_block),
```
MatthieuDartiailh, line 449: You should add a case using directly from_rs_block to catch the error on line 804.

spevs: Done.

pyvisa/testsuite/test_util.py
```diff
@@ -448,9 +558,14 @@ def test_weird_binary_block_header_raise(self):
     def test_binary_block_shorter_than_advertized(self):
         values = list(range(99))
         for header, tb, fb in zip(
-            ("ieee", "hp"),
-            (util.to_ieee_block, util.to_hp_block),
-            (util.from_ieee_block, util.from_hp_block),
+            ("ieee", "hp", "rs"),
```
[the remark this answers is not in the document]
spevs: Whoops :)

PR #552, opened 2020-10-01
Gpib commands

MatthieuDartiailh:
- [x] Executed ``black . && isort -c . && flake8`` with no errors
- [ ] Documented in docs/ as appropriate
- [ ] Added an entry to the CHANGES file

Add a more convenient interface for accessing GPIB commands and improve grou_execute_trigger.

@greyltc if you can test this it would be amazing.

PR #930, opened 2025-08-07
Fix group execute trigger

### LINKED ISSUE #929: group_execute_trigger does not work
`group_execute_trigger` fails with error `int too big to convert`

Minimal code to reproduce:
```
import pyvisa
rm=pyvisa.ResourceManager()
controller=rm.open("")
instr=rm.open_resource("GPIB0::10::INSTR")
controller = rm.open_resource("GPIB0::INTFC")
controller.group_execute_trigger(instr)
```

nothing should happen, but instead it returns `int too big to convert`

The problem is in file pyvisa\resources\gpib.py. I will be sending a PR.

hjf:
I've noticed the implementation has quite a few serious bugs. I've fixed the following:

- UNT and UNL commands were reversed. Source: [here](https://www.ni.com/docs/en-US/bundle/gpib-for-labview-nxg/page/ieee-488-command-messages.html).
- `talker` and `listener` static methods use 40 and 20 decimal as base address. It should be 40 and 20 hex.
- `secondary_address` method doesn't handle the 65535 (-1?) value set by default

- [x] Closes #929
- [ ] Added an entry to the CHANGES file

### DIFF
--- a/pyvisa/resources/gpib.py
+++ b/pyvisa/resources/gpib.py
@@ -60,22 +60,22 @@ class GPIBCommand(bytes, Enum):
     configure_enable = CFE = b"\x1f"
 
     #: UNT: UNTALK
-    untalk = UNT = b"\x3f"
+    untalk = UNT = b"\x5f"
 
     #: UNL: UNLISTEN
-    unlisten = UNL = b"\x5f"
+    unlisten = UNL = b"\x3f"
 
     @staticmethod
     def talker(board_pad) -> bytes:
         """MTA: MY TALK ADDRESS."""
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh (commented):
Thanks a lot for your contribution !!!

Could you reference the NI page directly in the code and maybe even copy it as part of the dosctring ? I guessed I missed this reference when I worked on #552.
Regarding MSA, I think I was under the impression that they started at 1 and may have used 95 because of that. Could you fix it to 96 as part of your PR even if you cannot test ?

MatthieuDartiailh:
Thanks

PR #986, opened 2026-07-07
Fix outdated manual install instructions in README

### LINKED ISSUE #839: Manual installation from source - lack of setup.py
Due to safety restrictions I have to install pyvisa manually from source. PIP is strictly forbidden. 
According to README:

"download and unzip the source distribution file and:
`    $ python setup.py install`"

But in the repository there is no setup.py file. 

To Reproduce
--------------

**Output of `pyvisa-info`**

hmms:
Fixes #839.

The README told users to run `python setup.py install` after downloading the source, but the project has moved to a `pyproject.toml`-based build and no longer ships a `setup.py`, so that command fails. The `easy_install` alternative listed alongside it was also dead: `easy_install` was removed from setuptools years ago and no longer exists as a command.

This replaces both with `pip install .` run from the unzipped source directory, and adds a note on installing fully offline via `--no-build-isolation`/`--no-index --find-links` (linking pip's local-project-installs docs), which is the actual constraint the reporter has: PyPI access is blocked by their organization's policy.

This is my first contribution to pyvisa, happy to adjust wording/scope per your preference.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@ PyVISA Changelog
 1.17 (unreleased)
 -------------------
 - add support for context manager protocol to ResourceManager PR #985
+- fix outdated manual install instructions in README PR #986
 
 1.16.2 (29-01-2026)
 -------------------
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Thanks

